PR #7, opened 2024-03-14
add dotenv for loading .env

es92:
Loading .env didn't work for me by default, added this for that to work (though maybe there is some other preferred way to load .env)


Onurbon:
good catch, thanks!

PR #9, opened 2024-04-10
support Claude  

Onurbon:
@colleenm I haven't tested this yet as I'm waiting to get an anthropic key. 

Overview: 
- I added a new option field `model`
- if `model` starts with `"gpt-"` we use OpenAI's API
- if `model` starts with `"claude-"` we use Anthropic's API
- default model remains gpt-4-turbo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules/
 .DS_Store
 google-credentials.json
 .env*
+.vscode/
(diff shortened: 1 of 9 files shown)


brittneygallagher:
@Onurbon I sent it to you via signal

Brandon-Perry (approved):

PR #39, opened 2024-12-04
first pass local pyserver

staceysv:
adds a basic Python server for the T3C LLM pipeline:
* Step 1: **POST /topic_tree/** : given a list of comments [1] return a tree of topics and subtopics
* Step 2: **POST /claims/**: given a topic tree and one comment at a time [2], extract claims from each comment and return a topic tree with claims placed correctly (including original quote) 
* Step 3: **PUT /sort_claims_tree/**: given a tree with placed claims [3] deduplicate the claims in each subtopic and return a sorted version of the full tree, where most common topics, subtopics, and claims each appear first in their parent branch

TODOS

P0
- [ ] how do we call this properly from JS/FE?
- [ ] how do we package this to run a Python backend in prod?
- [ ] check for headers/cors/user authentication gotchas
- [ ] correctly set up & pass around OpenAI and wandb API keys, esp for any external users
- [ ] set up & test client overrides of prompts/models

P1
- [ ] more serious load-testing with larger CSV files
- [ ] preferred/best practice for HTTP/API param/request style & more robust error checking (e.g type/max length) on FastAPI routes/body/queries/etc

P2
- [ ] proper pytest setup
- [ ] add cost logging back in, or build cost calculation around wandb
- [ ] add Weave back in if we miss it more than once
- [ ] style/format for Python code :), black/poetry?

[1]: format/schema for comment list:
```
class Comment(BaseModel):
  text: str
class CommentList(BaseModel):
  comments: List[Comment]
```
[2] format/schema for topic tree with comments
```
class CommentTopicTree(BaseModel):
  tree: dict
  comments: List[Comment]
```
[3] format/schema for claims tree
```
class ClaimTree(BaseModel):
  tree: dict
```

### DIFF
--- a/pyserver/main.py
+++ b/pyserver/main.py
@@ -24,6 +24,14 @@ class CommentTopicTree(BaseModel):
 class ClaimTree(BaseModel):
   tree: dict 
 
+# allow client to override model & all prompts
+class ClientLLMConfig(BaseModel):
+  model: str
+  system_prompt: str
+  comment_to_tree_prompt: str
+  comment_to_claims_prompt: str
+  claim_dedup_prompt: str 
+
 app = FastAPI()
 
 @app.get("/")
@@ -41,6 +49,7 @@ def comments_to_tree(comments: CommentList, log_to_wandb:bool = False):
   """
   client = OpenAI()
 
+  # TODO: client overrides of prompt!
   # append comments to prompt
   full_prompt = config.COMMENT_TO_TREE_PROMPT
   for comment in comments.comments:
@@ -64,7 +73,7 @@ def comments_to_tree(comments: CommentList, log_to_wandb:bool = False):
   try:
     tree = json.loads(response.choices[0].message.content)
   except:
-    # TODO: return an error?
+    print("Step 1: no topic tree: ", response)
     tree = {}
   usage = response.usage
     
@@ -130,7 +139,11 @@ def comment_to_claims(comment:str, tree:dict)-> dict:
     temperature = 0.0,
     response_format = {"type": "json_object"}
   )
-  claims = response.choices[0].message.content
+  try:
+    claims = response.choices[0].message.content
+  except:
+    print("Step 2: no response: ", response)
+    claims = {}
   return {"claims" : json.loads(claims), "usage" : response.usage}
 
 ####################################
@@ -149,7 +162,11 @@ def all_comments_to_claims(tree:CommentTopicTree, log_to_wandb:bool = False) -> 
   # TODO: batch this so we're not sending the tree each time
   for comment in tree.comments: 
     response = comment_to_claims(comment.text, tree.tree)
-    claims = response["claims"]
+    try:
+      claims = response["claims"]
+    except:
+      print("Step 2: no claims for comment: ", response)
+      claims = None
     # reference format
     #{'claims': [{'claim': 'Dogs are superior pets.', 'quote': 'dogs are great', 'topicName': 'Pets', 'subtopicName': 'Dogs'}]} 
     usage = response["usage"]
@@ -228,7 +245,11 @@ def dedup_claims(claims:list)-> dict:
     temperature = 0.0,
     response_format = {"type": "json_object"}
   )
-  deduped_claims = response.choices[0].message.content
+  try:
+    deduped_claims = response.choices[0].message.content
+  except:
+    print("Step 3: no deduped claims: ", response)
+    deduped_claims = {}
   return {"dedup_claims" : json.loads(deduped_claims), "usage" : response.usage}
 
 #####################################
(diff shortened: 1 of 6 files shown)


pyserver/main.py
```diff
@@ -0,0 +1,362 @@
...
+  )
+  deduped_claims = response.choices[0].message.content
+  return {"dedup_claims" : json.loads(deduped_claims), "usage" : response.usage}
+
+#####################################
+# Step 3: Sort & deduplicate claims #
+#-----------------------------------#
+@app.put("/sort_claims_tree/")
+def sort_claims_tree(claims_tree:ClaimTree, sort_type : str = "deduplicate", log_to_wandb: bool = False)-> dict:
+  #TODO: do we ever want to deduplicate without sorting? or sort without deduplicating?
```
colleenm, line 239: probably not.

eventually we'll want to sort by number of people who mentioned a topic instead of number of claims, but probably still fine to keep that in the same function

pyserver/main.py
```diff
@@ -0,0 +1,362 @@
...
+    })
+  return {"tree" : full_sort_tree, "per_topic_dupes" : nested_claims, "dupe_counts" : dupe_counts}
+
+#TODO: refactor into separate testing script
+
+######################
+# Testing the server #
+#--------------------#
+
+sample_tree_4o = {'tree': {'taxonomy': [{'topicName': 'Pets', 'topicShortDescription': 'General opinions about common household pets.', 'subtopics': [{'subtopicName': 'Cats', 'subtopicShortDescription': 'Positive sentiments towards cats.'}, {'subtopicName': 'Dogs', 'subtopicShortDescription': 'Positive sentiments towards dogs.'}, {'subtopicName': 'Birds', 'subtopicShortDescription': 'Uncertainty or mixed feelings about birds.'}]}]}, 'usage': {'completion_tokens': 125, 'prompt_tokens': 220, 'total_tokens': 345, 'completion_tokens_details': {'accepted_prediction_tokens': 0, 'audio_tokens': 0, 'reasoning_tokens': 0, 'rejected_prediction_tokens': 0}, 'prompt_tokens_details': {'audio_tokens': 0, 'cached_tokens': 0}}}
```
colleenm, line 340: accidental duplicate of test_pipeline.py? or intentionally also here?

pyserver/main.py
```diff
@@ -0,0 +1,362 @@
...
+              if len(claim_vals) > 0:
+                has_dupes = True
+                # extract relative index
+                claim_id = int(claim_key.split("Id")[1])
+                dupe_keys = [int(dupe_claim_key.split("Id")[1]) for dupe_claim_key in claim_vals]
+                dupe_counts[subtopic_data["claims"][claim_id]] = dupe_keys
+
+          # for logging to wandb
+          if log_to_wandb:
+            dupe_logs.append(["\n".join(subtopic_data["claims"]), json.dumps(deduped_claims, indent=1)])
```
colleenm, line 270: weird indent?

pyserver/main.py
```diff
@@ -0,0 +1,362 @@
...
+
+          if has_dupes:
+            nested_claims[subtopic] = {"dupes" : deduped_claims, "og" : subtopic_data["claims"]}
```
colleenm, line 273: I'm a little brain foggy and not sure I'm following the logic here right, but does this actually produce exactly the same schema we're currently sending to the client?

might be helpful at this point to explicitly describe the shape of the data blob the pipeline returns -- will also help keep things in sync with the client going forward (e.g. after Brandon's updates to fix the missing-quote-attribution issue)

colleenm (approved):

pyserver/main.py
```diff
@@ -0,0 +1,362 @@
...
+#---------------------------------#
+@app.post("/topic_tree/")
+def comments_to_tree(comments: CommentList, log_to_wandb:bool = False):
+  """
+  Given the full list of comments, return the tree of topics and subtopics
+  """
+  client = OpenAI()
+
+  # append comments to prompt
+  full_prompt = config.COMMENT_TO_TREE_PROMPT
```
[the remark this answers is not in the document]
staceysv: agree this is a bit confusing—`config` here is the name of a python module (`config.py`) which contains the default prompts and the default model name. I tried calling this `default_prompts.py` and loading in as `def_prompt.COMMENT_TO_TREE_PROMPT`, and that made `def_prompt.MODEL` more confusing to me. Open to other name suggestions for this Python module of config (better than `prompts.txt` I realized, doesn't need loading, and we can put more complicated/nested model definitions in there easily, e.g for OpenAI vs Claude).

In the future I expect to load defaults from the python server like this first, then check if the client has any overrides, and add them on top. Will add a test/TODO for this!

pyserver/main.py
staceysv: intentional for now—pytest only CHECKS, and doesn't easily SHOW output; need to set up something in parallel that does just show the output for a bunch of calls

pyserver/main.py
staceysv: correct logic but confusing yes—moved inside of "has_dupes", thank you!

pyserver/main.py
staceysv: yes this is only approximate for now — will test!

PR #40, opened 2024-12-08
LLM deduplication: cleaner docs, output format matches current JS

staceysv:
- deduplication step return format now perfectly matches current expected format of JS front-end
- much more detailed docstring with examples on Step 3 / deduplication & tree sorting step
- added a full test for all three steps — keeping this inside pyserver for interactive testing for now

### DIFF
--- a/pyserver/main.py
+++ b/pyserver/main.py
@@ -193,11 +193,11 @@ def all_comments_to_claims(tree:CommentTopicTree, log_to_wandb:bool = False) -> 
         if "subtopicName" in claim:
           if claim["subtopicName"] in node_counts[claim["topicName"]]["subtopics"]:
             node_counts[claim["topicName"]]["subtopics"][claim["subtopicName"]]["total"] += 1
-            node_counts[claim["topicName"]]["subtopics"][claim["subtopicName"]]["claims"].append(claim["claim"])
+            node_counts[claim["topicName"]]["subtopics"][claim["subtopicName"]]["claims"].append(claim)
           else:
-            node_counts[claim["topicName"]]["subtopics"][claim["subtopicName"]] = { "total" : 1, "claims" : [claim["claim"]]}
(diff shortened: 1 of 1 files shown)


pyserver/main.py
```diff
@@ -343,24 +532,34 @@ def sort_claims_tree(claims_tree:ClaimTree, sort_type : str = "deduplicate", log
...
+  net_usage = {"total_tokens" : TK_TOT,
+               "prompt_tokens" : TK_IN,
+               "completion_tokens" : TK_OUT}
+
+  return {"tree" : full_sort_tree, "usage" : net_usage} 
+# TODO: RETURN USAGE DUH
 
 #TODO: refactor into separate testing script
 
 ######################
```
Brandon-Perry, line 547: I think testing should be moved into a separate file.

Brandon-Perry (approved):

pyserver/main.py
staceysv: yes absolutely, there's a todo for that right below, and there's a pytest set up with test_pipeline.py—started making those tests work, realized it's all around equality assertions and we primarily want to test LLM calls that may not be entirely deterministic—will keep iterating

pyserver/main.py
```diff
@@ -43,24 +59,96 @@ def read_root():
 # Step 1: Comments to Topic Tree  #
 #---------------------------------#
 @app.post("/topic_tree/")
-def comments_to_tree(comments: CommentList, log_to_wandb:bool = False):
+def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:bool = False):
```
colleenm, line 62: maybe this is a python idiom I'm not familiar with, but why is this param called "req"? because it's data from the API request sent from the ts server?

I don't need it changed if it's clear to others, just want clarity

colleenm (approved):
looks great

PR #42, opened 2024-12-26
Pyserver add comment

Brandon-Perry:
Adds an id to comment and a commentId field to claims. 

Changes the doc strings as well to reflect these changes.

### DIFF
--- a/pyserver/main.py
+++ b/pyserver/main.py
@@ -82,15 +82,15 @@ def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:bool = False):
     },
     "comments": [
         {
-            "id: "..",
+            "id: "c1",
             "text": "I love cats"
         },
         {
-            "id: "..",
+            "id: "c2",
             "text": "dogs are great"
         },
         {
-            "id: "..",
+            "id: "c3",
             "text": "I'm not sure about birds"
         }
     ]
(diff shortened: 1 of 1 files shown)


pyserver/main.py
```diff
@@ -81,12 +82,15 @@ def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:bool = False):
     },
     "comments": [
         {
+            "id: "..",
```
staceysv, line 85: nit: prefer literal example in comments, so "id" : "123", then "id" : "456", "id" : "789". Can even put in an example like "id" : "my-random-string-here" to make it super clear to the reader. Also "id" is missing the closing quote here in all the rows.

pyserver/main.py
```diff
@@ -312,6 +319,7 @@ def all_comments_to_claims(req:CommentTopicTree, log_to_wandb:bool = False) -> d
                     "claims": [
                         {
                             "claim": "Cats are the best household pets.",
+                            commentId":"..",
```
staceysv, line 322: same minor nit: commentId missing starting quote, would prefer literal example of a comment id string

pyserver/main.py
```diff
@@ -585,6 +606,8 @@ def sort_claims_tree(claims_tree:ClaimTree, log_to_wandb: bool = False)-> dict:
                         "claims": [
                             {
                                 "claim": "Dogs are superior pets.",
+                                "commentId" "..",
```
staceysv, line 609: accidental double add

staceysv (approved):
looks great, thanks for adding this! Let's merge & can clean up comments when we get a chance since it's style/readability.

I prefer literal examples when possible for maximum informativeness/readability — so "id" : "1234", "id" : "my-random-hash-string", "id" : "auhfe7fh3er2", "commentId" : "12831724" etc. Don't have to be so random/can absolutely be repeated, and imho much more clear than "..", especially when an id can often have numerical, OR string, or some fancy/specific/reserved type (which we may want later to be, say, globally unique in our system, or correctly retrievable from some DB).

PR #46, opened 2024-12-30
Create report form [merge #44 first]

Brandon-Perry:
Based from #44 

Creates a new front-end for making reports


staceysv:
love the cost estimates especially—excellent work @Brandon-Perry !

PR #48, opened 2025-01-02
[WIP] pyserver cleanup

staceysv:
* standardize all the keys in API responses to `"data"` instead of various different strings
* separate and much cleaner test file :)
* minor fix in managing commentIds

Note: deduplication has gotten worse/changed for me from past tests—not sure what exactly could be happening here yet :)


objectivevan:
Looking into dedup

PR #49, opened 2025-01-09
Shared pyserve integration

Brandon-Perry:
PR to track our integration

Set to draft until we decide we're ready to merge

PR #51, opened 2025-01-09
#49 << Pyserver fix

Brandon-Perry:
Merges into #49 

I figured out the issue that I showed @staceysv earlier. Turns out it was due to usage not being in JSON format. I added .model_dump() to it and that fixed the issue.


staceysv (commented):
thanks for finding the fix!! I would like to figure out what's happening here/why it's different from the other routes—I don't think we should need this.

staceysv (approved):

staceysv:
I still want to figure out what's happening here, and merging for now to see if this branch will work for me

PR #54, opened 2025-01-09
next client env parsing

Brandon-Perry:
Redoes env management on the nextjs client

Now we parse the env when the Next server starts and tells you when you fuck up in an easily readable message.
Also exports the validated env as a var that can be passed around on the Next server. This way we don't have to do existential checks whenever we want to use one.

Example error message:
[Error: There are error(s) with your next-js env: 
 
0) PIPELINE_EXPRESS_URL env var is missing. This is a required. 
]


Brandon-Perry:
@es92 is this a secure approach? the import 'server-only' at the top should prevent it from being passed to the client. But idk if there's any issues with having it saved as a const.

objectivevan:
we should have a conversation with C and S about our threat models re: highly sophisticated attackers, but this seems sane and good enough for right now.

PR #56, opened 2025-01-09
Pyserver pass API key

Brandon-Perry:
The pyserver api now takes an api key to use for ChatGPT. 

Changes the shape of the third route's parameters so it can take an llm config.

Did a successful full pipeline test, so it should be working.

@staceysv can you make sure your tests still work with this? Idk how you manage your tests and such.


staceysv:
looks good, made a TODO of re-checking tests

PR #65, opened 2025-01-13
updated example env

Brandon-Perry:
Updated example envs


express-server/.example-env
```diff
@@ -1,9 +1,10 @@
...
+export OPENAI_API_KEY=
+export GCLOUD_STORAGE_BUCKET= name of your bucket
+export CLIENT_BASE_URL= for dev: http://localhost:3000
+export GOOGLE_CREDENTIALS_ENCODED= base 64 encoding of your credentials
+export PYSERVER_URL= for dev: http://localhost:8000
+export FIREBASE_DATABASE_URL= found in your firebase project
+export REDIS_HOST= for dev: localhost
+export REDIS_PORT= for dev: 6379
+export GOOGLE_APPLICATION_CREDENTIALS= path to your credentials. Suggest using ./credentials
+export NODE_ENV=
```
staceysv, line 10: NODE_ENV: is this "dev" vs "prod"? can we add that?

next-client/.example-env.local
```diff
@@ -1,2 +1,9 @@
 # How env works on NextJS: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables
-PIPELINE_EXPRESS_URL= # This is by default localhost:8080 on dev
\ No newline at end of file
+PIPELINE_EXPRESS_URL= # This is by default localhost:8080 on dev
+# Firebase keys below should be found on your firebase project. These are not sensitive and can be shared with the client.
```
staceysv, line 3: nice, is there a particular page which shows these? like "under main menu > project settings in your Firebase admin account"?

staceysv (approved):
this is really helpful, thank you! merging now, would love to add the details from the two comments (how is NODE_ENV used and can we tell the user where in Firebase to find all the config info we need, assuming it might not be super obvious).

Also, would be great to have a single AOI-local-dev-mode env file we can reuse (perhaps except from 1-2 fields to be overwritten), instead of everyone individually filling out these fields—will add to the Runbook.

PR #71, opened 2025-01-17
cruxes: compute and log controversy matrix

staceysv:
EDIT: this is ready to merge now!

This branch adds a pyserver route `cruxes` to generate crux claims for each subtopic via an LLM call (to formulate a statement which would best split the speakers into "agree" and "disagree" camps). Tested on a few small examples + logging to W&B correctly. Note, this already includes the speaker logic from the `people_atr` branch

TODOs:
- [x] clean up last logged Table(s): what do we want to visualize for each crux?
- LATER how do we best generate a matplotlib plot & return an image file in FastAPI?
- LATER reconsider default workspace layout in W&B: do we want token counts or routes as primary keys?


staceysv:
ready to merge before it gets any bigger! very open to eventual TS improvements, and this works for now

outstanding items:

- [ ] document the route: better in code & more externally in readme
- [ ] clean example or two in W&B report with clean dashboards
- [ ] add a UI element for editing crux instructions
- [ ] set top_K and cruxInstructions in prompts.json in common/
- [ ] more example datasets
- [ ] how to efficiently compute cruxes for older reports
- [ ] improve crux extraction prompt & calculations (research :)

express-server/src/workers.ts
```diff
@@ -104,6 +112,22 @@ const setupPipelineWorker = (connection: Redis) => {
         llm: claimsLLMConfig,
       });
 
+      console.log("Step 2.5: Optionally extract cruxes");
```
Brandon-Perry, line 115: Since this step is optional, there should be a way for this part to not run?

Brandon-Perry (changes requested):

common/apiPyserver.ts
```diff
@@ -210,3 +214,49 @@ export const sortClaimsTreeResponse = z.object({
...
+ * - basic details of an extractec cruxClaim: LLM-generated crux claim,
+ *   lists of speakers who agree/disagree, LLM-generated explanation
+ * - controversy matrix: numeric scores for the cross-product of crux claims,
+ *   higher score is more controversial
+ * - human-readable cruxes: top K scores and corresponding crux pairs from the
+ *   controversy matrix
+ * - one cruxDetails object to package this together
+ */
+
+const cruxClaim = z.object({
```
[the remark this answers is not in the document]
staceysv: ah great point, thank you — looking into how to resolve the duplicates! Would be great to keep these more organized in general.

express-server/src/workers.ts
staceysv: yeah that's a good question! there is NOT right now, it will just run by default—would be nice to add one in the form input — e.g. ask users to check a box if they want cruxes, and also optionally edit the crux prompt. I have not gotten around to making this form input/optional step, so I think it needs to be default-on for now/today.

Brandon-Perry (approved):

PR #73, opened 2025-01-18
Videos in reports

Brandon-Perry:
Adds the ability to handle video sources and display them in the report.

You should be able to see it in action by spinning up a dev instance and going to: 
`http://localhost:3000/report/https%3A%2F%2Fstorage.googleapis.com%2Ftttc-light-dev%2Ftest%2Ftest5.json`

This will show the Heal-Michigan videos.


common/morphisms/pipeline.ts
```diff
@@ -210,21 +247,21 @@ export const getReportDataObj = (
...
-    total: 0,
-  },
-  costPerToken: {
-    denomination: "$",
-    value: 0,
-  },
-  model: "claude-instant-v1",
-  instructions: "",
-};
+// const buildStageData: schema.PipelineStepData = {
```
colleenm, line 250: @Brandon-Perry can you add a comment about why this is left in but commented out, if keeping it isn't an error

colleenm:
@Brandon-Perry this is beautiful!!!

quick UI nit I think should be easy to fix: it's sometimes difficult to move from hovering over a claim to interacting with the popup, but I think if the popup were a few pixels closer to the cursor, this behavior would actually work just fine.

estimate: having the lower edge of the popup cover 25-50% of the claim square it's associated with would work, because then there's no danger of the cursor leaving the claim square to move into the modal.

example at ~50%:
Screen Shot 2025-01-28 at 10 48 45
Screen Shot 2025-01-28 at 10 49 17

Brandon-Perry:
@colleenm Alright try that now

PR #78, opened 2025-01-30
fix venv and condense suggestions across different readmes

staceysv:
- corrects venv to .venv
- removes some old instructions
- attempts to clarify about google credentials


colleenm:
lgtm

PR #80, opened 2025-02-01
Add Apache 2.0 license to repo

colleenm:


colleenm:
Made the decision to go with Apache last year (some opportunities were contingent on it) -- actually adding it now

PR #82, opened 2025-02-03
NextJS build uses tsconfig.build.json file now

Brandon-Perry:
Updates NextJS build so that it uses tsconfig.build.json instead. This helps in case we run into any type errors in tests and stories, so it doesn't cause the whole thing to crash, and they aren't bundled. 

Making this a draft for now to not mess with current devops stuff. But let me know when it would be fine to merge it - it'll probably save devops some headache.

### DIFF
--- a/next-client/next.config.js
+++ b/next-client/next.config.js
@@ -3,6 +3,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   /* config options here */
+  typescript: {
+    tsconfigPath: "tsconfig.build.json",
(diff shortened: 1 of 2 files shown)


staceysv (approved):

staceysv:
@Brandon-Perry this is from a while ago, shall we merge this?

Brandon-Perry:
@staceysv I'll need to review it again, but in principle yes

justinstimatze (approved):
LGTM

PR #85, opened 2025-02-04
Add duplicates to get n people

Brandon-Perry:


staceysv (approved):
brilliant, thank you!! really appreciate the test pattern as well, let's do more of these!

PR #86, opened 2025-02-05
logging, json, layout

staceysv:
- cleaner wandb logging, easier to configure for local-mode-only
- json loading in try-except — see if this is part of the trouble with some csvs
- tiny layout fix to change text on tab in browser

### DIFF
--- a/next-client/src/app/layout.tsx
+++ b/next-client/src/app/layout.tsx
@@ -7,7 +7,7 @@ import { getAuthenticatedAppForUser } from "@src/lib/firebase/serverApp";
 export const dynamic = "force-dynamic";
 
 export const metadata = {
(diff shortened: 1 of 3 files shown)


next-client/src/app/layout.tsx
```diff
@@ -7,7 +7,7 @@ import { getAuthenticatedAppForUser } from "@src/lib/firebase/serverApp";
 export const dynamic = "force-dynamic";
 
 export const metadata = {
-  title: "Talk the City",
+  title: "Talk to the City",
```
Brandon-Perry, line 10: nice catch!

Brandon-Perry (approved):
lgtm

PR #87, opened 2025-02-06
Deployment docker compose rebased

Brandon-Perry:
Rebased version of deployment-docker-compose

Has a docker-compose setup + some fixes to issues we found during deployment.


staceysv (approved):
amazing work, thank you!
noting a few TODOs/questions for later
- why do we have a duplicate main loop in two spots in main.py?
- how can we have the easiest swap-between-local-and-prod for Redis so we don't get confused/forget?
- what's the more robust way to fix ReportProgress so our default status is not an error?

PR #88, opened 2025-02-06
hash unique report titles via sha256

staceysv:
- for any new report, take current unique filename (defined in utils)
- hash it via sha256 implementation from Evan
- return that as much longer but much harder to guess unique url string
- bucket names are still visible but report ids are not guessable for publicly visible report urls


Brandon-Perry:
@objectivevan fyi I didn't get a chance to review this before merging - I would wait until everyone gets a chance to submit their review first.

PR #95, opened 2025-02-08
Hide hamburger menu button, and add buttons to header, until we have a menu that the hamburger button shows

colleenm:
Quick fix to get mobile closer to parity

### DIFF
--- a/next-client/src/components/navbar/Navbar.tsx
+++ b/next-client/src/components/navbar/Navbar.tsx
@@ -19,13 +19,16 @@ function Navbar({ currentUser }: { currentUser: User | null }) {
       <Headline />
       <Row
         gap={2}
-        className="items-center content-center md:justify-self-end hidden sm:flex"
+        className="items-center content-center md:justify-self-end sm:flex"
       >
-        <RepoGithub />
+        <Row
+          gap={2}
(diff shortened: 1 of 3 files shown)


next-client/src/components/navbar/components/LoginButton.tsx
```diff
@@ -72,7 +72,7 @@ export default function LoginButton({
   if (loading) {
     return (
       <div>
-        <p>Loading...</p>
+        <Button disabled={true}>...</Button>
```
colleenm, line 75: I may have gotten a little carried away with UI fixes, can revert this if people don't like it

next-client/src/components/navbar/Navbar.tsx
```diff
@@ -19,14 +19,17 @@ function Navbar({ currentUser }: { currentUser: User | null }) {
...
+        <Row
+          gap={2}
+          className="items-center content-center md:justify-self-end hidden sm:flex"
+        >
+          <RepoGithub />
+        </Row>
 
         <About />
 
-        <LanguageSelector />
```
colleenm, line 28: Can add this back in once it does things

Brandon-Perry (approved):

PR #99, opened 2025-02-09
move OAI keys to headers

staceysv:
- changes pyserver routes to take OAI api keys as headers instead of as request param in the json body
- changes TS calling accordingly (@Brandon-Perry sincere apologies for style/convention misses here—was only focused on getting it to work)
- pyserver routes crash early if not authenticated
- now we don't pass around a key in the body which might accidentally be logged
- may add extra middleware per Portia's recommendations and wanted to start a placeholder PR
- in the future, will want to consider how this extends to multiple keys, e.g. calling Claude instead

### DIFF
--- a/express-server/src/pipeline/claimsStep.ts
+++ b/express-server/src/pipeline/claimsStep.ts
@@ -11,7 +11,7 @@ const typedFetch =
       body: JSON.stringify(bodySchema.parse(body) as z.infer<T>),
       headers: {
         "Content-Type": "application/json",
-        "openai-api-key" : openaiAPIKey
+        "openai-api-key": openaiAPIKey,
       },
     });
 
@@ -24,7 +24,11 @@ const logger =
     return arg;
   };
 
-export async function claimsPipelineStep(env: Env, openaiAPIKey: string, input: ClaimsStep["data"]) {
+export async function claimsPipelineStep(
+  env: Env,
+  openaiAPIKey: string,
+  input: ClaimsStep["data"],
+) {
   const { data, usage } = await pyserverFetchClaims(
     `${env.PYSERVER_URL}/claims`,
     input,
--- a/express-server/src/pipeline/claimsStep.ts
+++ b/express-server/src/pipeline/claimsStep.ts
@@ -5,12 +5,13 @@ import { z } from "zod";
 
 const typedFetch =
   <T extends z.ZodTypeAny>(bodySchema: T) =>
-  async (url: string, body: z.infer<T>) =>
+  async (url: string, body: z.infer<T>, openaiAPIKey: string) =>
     await fetch(url, {
       method: "POST",
       body: JSON.stringify(bodySchema.parse(body) as z.infer<T>),
       headers: {
         "Content-Type": "application/json",
+        "openai-api-key" : openaiAPIKey
       },
     });
 
@@ -23,10 +24,11 @@ const logger =
     return arg;
   };
 
-export async function claimsPipelineStep(env: Env, input: ClaimsStep["data"]) {
+export async function claimsPipelineStep(env: Env, openaiAPIKey: string, input: ClaimsStep["data"]) {
   const { data, usage } = await pyserverFetchClaims(
     `${env.PYSERVER_URL}/claims`,
     input,
+    openaiAPIKey,
   )
     .then((res) => res.json())
     .then(logger("claims step returns: "))
--- a/express-server/src/pipeline/topicTreeStep.ts
+++ b/express-server/src/pipeline/topicTreeStep.ts
@@ -11,7 +11,7 @@ const typedFetch =
       body: JSON.stringify(bodySchema.parse(body) as z.infer<T>),
       headers: {
         "Content-Type": "application/json",
-        "openai-api-key" : openaiAPIKey
+        "openai-api-key": openaiAPIKey,
       },
     });
 
@@ -32,7 +32,7 @@ export async function topicTreePipelineStep(
   const { data, usage } = await pyserverFetchTopicTree(
     `${env.PYSERVER_URL}/topic_tree`,
     input,
-    openaiAPIKey
+    openaiAPIKey,
   )
     .then((res) => res.json())
     .then(logger("topic tree step returns: "))
--- a/express-server/src/pipeline/topicTreeStep.ts
+++ b/express-server/src/pipeline/topicTreeStep.ts
@@ -5,12 +5,13 @@ import { z } from "zod";
 
 const typedFetch =
   <T extends z.ZodTypeAny>(bodySchema: T) =>
-  async (url: string, body: z.infer<T>) =>
+  async (url: string, body: z.infer<T>, openaiAPIKey: string) =>
     await fetch(url, {
       method: "POST",
       body: JSON.stringify(bodySchema.parse(body) as z.infer<T>),
       headers: {
         "Content-Type": "application/json",
+        "openai-api-key" : openaiAPIKey
       },
     });
 
@@ -25,11 +26,13 @@ const logger =
 
 export async function topicTreePipelineStep(
   env: Env,
+  openaiAPIKey: string,
   input: TopicTreeStep["data"],
 ) {
   const { data, usage } = await pyserverFetchTopicTree(
     `${env.PYSERVER_URL}/topic_tree`,
     input,
+    openaiAPIKey
   )
     .then((res) => res.json())
     .then(logger("topic tree step returns: "))
--- a/pyserver/main.py
+++ b/pyserver/main.py
@@ -16,7 +16,8 @@ For local testing, load these from a config.py file
 import os
 import sys
 from pathlib import Path
-from fastapi import FastAPI
+from fastapi import Depends, FastAPI
+from fastapi.security import APIKeyHeader
 from openai import OpenAI
 from pydantic import BaseModel
 from typing import List, Union
@@ -30,7 +31,8 @@ sys.path.append(str(current_dir))
 import config
 from utils import cute_print
 
-app = FastAPI() 
+app = FastAPI()
+header_scheme = APIKeyHeader(name="openai-api-key") 
 
 @app.get("/")
 def read_root():
@@ -48,7 +50,6 @@ class LLMConfig(BaseModel):
   model_name: str
   system_prompt: str
   user_prompt: str
-  api_key: str
  
 class CommentsLLMConfig(BaseModel):
   comments: List[Comment]   
@@ -85,7 +86,7 @@ def read_root():
 # Step 1: Comments to Topic Tree  #
 #---------------------------------#
 @app.post("/topic_tree")
-def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:str = config.WANDB_GROUP_LOG_NAME) -> dict:
+def comments_to_tree(req: CommentsLLMConfig, api_key: str = Depends(header_scheme), log_to_wandb:str = config.WANDB_GROUP_LOG_NAME) -> dict:
   """
   Given the full list of comments, return a corresponding taxonomy of relevant topics and their
   subtopics, with a short description for each.
@@ -165,7 +166,8 @@ def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:str = config.WANDB_GRO
     }
   }
   """
-  api_key = req.llm.api_key
+  if not api_key:
+    raise HTTPException(status_code=401)
   client = OpenAI(
     api_key=api_key
   )
@@ -237,15 +239,10 @@ def comments_to_tree(req: CommentsLLMConfig, log_to_wandb:str = config.WANDB_GRO
   # choosing the latter for now 
   return {"data" : tree["taxonomy"], "usage" : usage.model_dump()}
 
-def comment_to_claims(llm:dict, comment:str, tree:dict)-> dict:
+def comment_to_claims(client, llm:dict, comment:str, tree:dict)-> dict:
   """
   Given a comment and the full taxonomy/topic tree for the report, extract one or more claims from the comment.
   """
-  api_key = llm.api_key
-  client = OpenAI(
-    api_key=api_key
-  )
-
   # add taxonomy and comment to prompt template
   taxonomy_string = json.dumps(tree)
    
@@ -285,7 +282,7 @@ def comment_to_claims(llm:dict, comment:str, tree:dict)-> dict:
 # Step 2: Extract and place claims #
 #----------------------------------#
 @app.post("/claims")
-def all_comments_to_claims(req:CommentTopicTree, log_to_wandb:str = config.WANDB_GROUP_LOG_NAME) -> dict:
+def all_comments_to_claims(req:CommentTopicTree, api_key: str = Depends(header_scheme), log_to_wandb:str = config.WANDB_GROUP_LOG_NAME) -> dict:
   """
   Given a comment and the taxonomy/topic tree for the report, extract one or more claims from the comment.
   Place each claim under the correct subtopic in the tree.
@@ -397,6 +394,11 @@ def all_comments_to_claims(req:CommentTopicTree, log_to_wandb:str = config.WANDB
     }
   }
   """
+  if not api_key:
+    raise HTTPException(status_code=401)
+  client = OpenAI(
+    api_key=api_key
+  )
   comms_to_claims = []
   comms_to_claims_html = []
   TK_2_IN = 0
@@ -406,7 +408,7 @@ def all_comments_to_claims(req:CommentTopicTree, log_to_wandb:str = config.WANDB
   node_counts = {}
   # TODO: batch this so we're not sending the tree each time
   for comment in req.comments:
-    response = comment_to_claims(req.llm, comment.text, req.tree)
+    response = comment_to_claims(client, req.llm, comment.text, req.tree)
     try:
        claims = response["claims"]
        for claim in claims["claims"]:
@@ -508,15 +510,10 @@ def all_comments_to_claims(req:CommentTopicTree, log_to_wandb:str = config.WANDB
 
   return {"data" : node_counts, "usage" : net_usage}
 
-def dedup_claims(claims:list, llm:LLMConfig)-> dict:
+def dedup_claims(client, claims:list, llm:LLMConfig)-> dict:
   """
   Given a list of claims for a given subtopic, identify which ones are near-duplicates
   """
-  api_key = llm.api_key
-  client = OpenAI(
-    api_key=api_key
-  )
-
   # add claims with enumerated ids (relative to this subtopic only)
   full_prompt = llm.user_prompt
   for i, orig_claim in enumerate(claims):
@@ -553,7 +550,7 @@ def dedup_claims(claims:list, llm:LLMConfig)-> dict:
 # Step 3: Sort & deduplicate claims #
 #-----------------------------------#
 @app.put("/sort_claims_tree/")
-def sort_claims_tree(req:ClaimTreeLLMConfig, log_to_wandb:str = config.WANDB_GROUP_LOG_NAME)-> dict:
+def sort_claims_tree(req:ClaimTreeLLMConfig, api_key: str = Depends(header_scheme), log_to_wandb:str = config.WANDB_GROUP_LOG_NAME)-> dict:
   """
   Sort the topic/subtopic tree so that the most popular claims, subtopics, and topics
   all appear first. Deduplicate claims within each subtopic so that any near-duplicates appear as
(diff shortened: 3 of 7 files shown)


staceysv:
added middleware! not sure if I'm testing it adequately though :) now wondering if we should also do something more explicit like this: https://www.uvicorn.org/settings/#https since the fastapi server is running on https :)

staceysv:
I've updated & tested this again! @portiabytes and @Brandon-Perry , I would love your take on this:
- moves OAI keys into headers across pyserver & TS requests
- adds tests for bad/missing headers
- adds middleware but doesn't enable it

Major TODOs/areas to figure out before merging:
1. What changes do we need in TS in order to enable the Pyserver/FastAPI security middleware?
2. How do we best test the resulting changes — across dev and prod? 
3. Option to share a single key: in the latest main, we share a single OpenAI API key as an env variable. I renamed this to test the headers change — as we switch the key field from body to headers, shall we:
A: pass this exact key from the env variable around in headers to all pyserver requests?
B: allow & use it if set globally in env, otherwise look in headers, or 
C: something else?

portiabytes:
This all looks great so far, def adds many needed changes. 

Further recs for this PR:

1. Complete the API key header implementation:
- Enable the security middleware in pyserver/main.py
- Add consistent header validation across all API endpoints
- Add rate limiting to the API key endpoints?
2. Add request validation
3. Add error handling
4. Enable CORS properly
5. Add request logging

I can create some example snippets to make these easy for you to include in this PR if you like.

Ongoing thoughts: The current changes are a fantastic start but need the below additions to be production-ready imo. 

Infrastructure Security:

- Add WAF configuration
- Implement proper key rotation
- Add monitoring and alerting
- Application Security:
- Add input sanitization
- Implement request throttling
- Add API key validation service

Monitoring:

- Add structured logging
- Implement error tracking
- Add performance monitoring

Questions:

- Should we add API key validation against OpenAI's API before accepting requests?
- What's the expected rate limiting strategy for user-provided API keys?
- Should we implement any usage tracking for billing/monitoring purposes? We probs want to know who is using t3c more or less, not sure what logging/metrics we have set up so far.
- What's the plan for handling API key rotation?

staceysv:
Awesome, thank you & great questions!

Completing this PR

some context:
- HTTP/S may not be the right long-term solution: I'd like to move to a message queue/Evan suggested websockets/something more robust & persistent than HTTP(S) for the longer reports soon (so the user isn't effectively refreshing a tab for 15 mins/having to rerun from scratch if one of many hundreds of requests errors) — would be great for this line of work to support that approach as well.
- sorting responses from highest priority for this PR to lowest

- **Enable CORS properly**:  yes please, I'm hoping this is some magic flags/params?

- **Enable the security middleware in pyserver/main.py:** yes, we should do this — I could use some help specifically on what the matching change/best practice is on the TS side. If I enable current FastAPI middleware, the TS side breaks/all TS requests to Pyserver need to be made HTTPS.

- **Add request validation**: if this means make sure it's a type string of max length N, yes please! Definitely want this to sanitize for basic jailbreaks, and/or potentially/optionally strip any PII, and slightly lower priority.

- **Add error handling** : would be nice to track the error codes in the logs easily! there's some basic handling in pyserver. Could do better esp with json/larger reports/dupes, also lower priority/I think the current version mostly does ok except for hard crashes of pipeline :)

- **Add request logging**: this could mean lots of things — I think it would be useful to trace/catalogue what we do log now and note if anything is extra (not leaking API keys!!!) or crucial & missing (when we get a 500, or hit a rate limit from some super malformed report). If we mean the contents/details of the request — right now some of this happens in pyserver / the TS components — can definitely be improved/standardized/ingested in better ways! the main type of logging I'm interested in is to W&B for evaluation/downstream analysis, and currently this is in dev-mode only. Happy to discuss details of what we'd like for that in the long term.

- **Add rate limiting to the API key endpoints?**: hmmmm, this is interesting. I think in the near term, we haven't needed to run large enough/concurrent enough reports to require this. I've done a tiny bit of this for a single report (lowered batch size from 10 to 2 in the original pipeline to avoid failures/rate limits). I suggest making sure we log the QPS per route in one clear place first/ASAP (Firebase...?), see what our typical numbers are, set up the scaffolding for a rate limit if we need, then be able to turn it on quickly?

- **Add consistent header validation across all API endpoints** : so far the headers are just json and keys — does this mean checking if the key is valid before sending a request? I'm ok to do this/we will need to do this more rigorously eventually if we're ever mixing models between calls. Also to the extent it slows things down, i think in the near term it's totally fine to validate once when the user hits "submit" —or even right after they enter their API key into the form — and then worst case let the request fail. Also lmk if this header validation refers to something else/more specific!

staceysv:
On additional/next steps: yes please/we need all of these!

Infrastructure Security:
- Add WAF configuration - yes please
- Implement proper key rotation - for which keys?
- Add monitoring and alerting - yes please/definitely — in FIrebase?

Application Security: yes, I would say before making dashboards of user counts :)

Monitoring: is this beyond / separate from the infra security monitoring & alerting? we definitely need this, I have it an "observability & production monitoring' bucket of work across GCP + Heroku, and would be GREAT to unify in one set of dashboards 🙏

staceysv:
Questions:
- Should we add API key validation against OpenAI's API before accepting requests? Yes, answered this above, I think we should do this once after user puts in their key into the form. And if/when we get to multiple keys, each time before a new key is used. I think if it's not a lot of time overhead, we could also do it for every request, and since the key doesn't change so far for the pipeline, it should just stay valid? We could potentially make things fancier, e.g. around estimating how much the report will cost/time it will take and helping the user decide on or configure that, and I would say lower priority.

- What's the expected rate limiting strategy for user-provided API keys?
Right now it's showing a rough estimate of the report cost and encouraging testing with smaller reports — users providing their own key at this early alpha stage know what they are getting into :) I think it would be great to provide more precise estimates, and start writing a plan in anticipation of more (than zero) folks using this simultaneously :)

- Should we implement any usage tracking for billing/monitoring purposes? Yes this fits into prodmon/obsevability work that we absolutely need, and I think needs to happen in Firebase in the short term — would love to chat more about this/for someone to take this.

Brandon-Perry:
Ran a report, no TS errors. Looks like everything is fine as far as I can tell. It's specifically when you comment back in the security middleware that's the issue right?

Brandon-Perry:
Are there any TS errors that are still an issue?

staceysv:
@Brandon-Perry thanks the specific questions/issues we need to figure out are below:
1. How do we get TS to work with these two changes (HSTS  in FastAPI / Pyserver? Currently if I set Pyserver NODE_ENV to "prod", or change these two check to be active "dev" — I get failures from TS because those requests are NOT using https.

- enable HSTS in prod
if os.getenv('NODE_ENV') == 'prod': response.headers['Strict-Transport-Security'] = 'max-age=31536000; includeSubDomains'

- enable strict HTTPS in prod (HTTP calls will fail)
if os.getenv('NODE_ENV') == 'prod': app.add_middleware(HTTPSRedirectMiddleware)

2. As you address that/if it DOES "just work" in prod — can you write up/explain what are the steps we need to take—whether locally or in a staged environment—so that we can properly test that the correct behavior actually happens in production?

staceysv:
Light editing to @portiabytes 's suggestions & additional thoughts

Adds a validation for env urls (CLIENT_BASE_URL and PYSERVER_URL) to be https in prod only.

P0

1. The current changes are in the definition of env in `express-server/src/types/context.ts`. There's also a `validateEnv()` function at the end of that file that is used throughout the repo. Should we 
a) make sure this check happens in that validateEnv function instead
b) make sure that `validateEnv()` function gets called EVERYWHERE we access process.env, and/or 
c) something else?

2. Actually testing this: we can push to prod or create a new staging instance and see if it works — I'm happy to do this/we need to do this eventually, AND if someone can be pretty sure about this being the right approach first/that it Should Just Work (TM) in prod, that would greatly help

P1

3. Do we need to add this check for EXPRESS_PIPELINE_URL also? perhaps in `next-client/src/server-env.ts` ? Should we have some other checks there?

4. There's a recommended change in claimsStep only, presumably to redirect any incoming HTTP requests to HTTPS if needed.
a) Is this actually the right interpretation?
b) Do we need this?
b) If we do need this, should we not add it to all the other steps (topicTreeStep, sortClaimsTree, etc)?

portiabytes:
Just pushed what I hope are the right changes here.

P0: 
I went with option A + B, which I think gives us the most maintainable approach:
- Moved the HTTPS validation logic from the Zod schema into the validateEnv() function
- Made sure validateEnv() is called early and consistently in any files that used to access process.env directly (like storage.ts)

open to thoughts, but this is what I see as benefit:
- central validation in one function that's easier to maintain and update
- clear separation between basic validation (is it a valid URL?) in the schema and production-specific validation (is it HTTPS?) in the func
- cleaner error msgs that explicitly state when we're rejecting a URL bc of missing HTTPS


I believe the approach is solid, but I agree we should def spin up and test in a staging environment to verify. This is a ton of changes to go out without some kind of validation in my mind, but I know I spent too long at fruitco and got very paranoid, so we might be more comfortable with blowing prod up once in a while, I'm just not excited about a first deploy being with my own relatively untested code.


P1: 
I added the same HTTPS check for PIPELINE_EXPRESS_URL in next-client/src/server-env.ts, makes our Next.js client also require HTTPS URLs for the Express server in production envs


Re your questions about the redirect option in claimsStep.ts:
a) Not quite. The redirect: "follow" option doesn't specifically handle HTTP-to-HTTPS redirects - it's a general setting that tells fetch to automatically follow any server redirects (which is actually the default behavior)
b) While technically not required (since "follow" is the default), I decided to keep it because:
- being explicit about redirect behavior makes the code easier to understand
- it ensures consistent behavior if fetch defaults ever change
- aligns with our general approach of being explicit about security-related settings
c) yeah, I standardize this now across all pipeline steps (claimsStep, topicTreeStep, and sortClaimsTree). all three now use the same pattern with a consistent comment explaining purpose


Let me know if you'd like me to make any other changes or want me to switch up the approach.

express-server/src/pipeline/topicTreeStep.ts
```diff
@@ -25,15 +35,18 @@ const logger =
...
   const { data, usage } = await pyserverFetchTopicTree(
     `${env.PYSERVER_URL}/topic_tree`,
     input,
+    openaiAPIKey,
+    env.NODE_ENV === "prod"
   )
     .then((res) => res.json())
     .then(logger("topic tree step returns: "))
-    .then(apiPyserver.topicTreeResponse.parse);
+    .then(apiPyserver.topicTreeReply.parse);
```
staceysv, line 49: oops this should stay as `topicTreeResponse` for now — or let's fix for consistency, we have apiPyserver.ts defining topicTreeResponse, claimsReply, sortClaimsTreeResponse, and cruxesResponse — maybe the min change is claimsResponse for consistency.

express-server/src/types/context.ts
```diff
@@ -24,21 +24,38 @@ class EnvValidationError extends Error {
...
     required_error: "Missing encoded GCloud credentials",
   }),
-  CLIENT_BASE_URL: z
-    .string({ required_error: "Missing CLIENT_BASE_URL" })
-    .url({ message: "PYSERVER_URL in env should be a valid url" }),
-  PYSERVER_URL: z
-    .string({ required_error: "Missing PYSERVER_URL" })
-    .url({ message: "PYSERVER_URL in env should be a valid url" }),
+  // Basic URL validation - additional HTTPS check for prod will be done in validateEnv
+  CLIENT_BASE_URL: z.string()
```
Brandon-Perry, line 34: What was wrong with the original parsers? Were they not checking correctly or something?

express-server/src/types/context.ts
```diff
@@ -50,24 +67,13 @@ export const env = z.object({
   FIREBASE_DATABASE_URL: z
     .string({ required_error: "Missing FIREBASE_DATABASE_URL" })
     .url({ message: "FIREBASE_DATABASE_URL in env should be a valid url" }),
-  // REDIS_HOST: z.string({ required_error: "Missing REDIS_HOST" }),
```
staceysv, line 53: there's a subtle dev/prod difference hiding here:
- in dev, we use REDIS_HOST + REDIS_PORT (pointing to a local process)
- in prod, we use REDIS_URL (pointing to a remote instance)
we can't get rid of this without accounting for both / merging these different config approaches

staceysv:
thank you for the updates @portiabytes ! looking much better overall, left a few specific notes / new list for us to address:
- [ ] REDIS HOST/PORT/URL: we need all three to make dev AND prod work — let's not remove them yet
- [ ] let's stick with Response instead of Reply for the returns of pipeline steps in apiPyserver.ts — change claimsReply to claimsResponse and topicTreeReply back to topicTreeResponse
- [ ] env check clarity / documentation: I appreciate the new functionality / separation of checking for valid urls vs https urls, and it looks like we have two separate phases of this in context.ts — one inside and one outside the validateEnv function, then a third way in server-env.ts, and another one in pyserver/main.py (this one inevitable, different language!). Is there any way to condense/simplify/unify this a bit? Have everything inside the function so we could, e.g., validateEnv in all three TS cases, perhaps with different input args like configureFor=prod or configurFor=dev? Or at the very least, a single comment/docstring/README to list our requirements for dev vs prod and what we check / where those checks happen in the code?
- [ ] I'm merging in main so I can test this — want to make sure the new cruxes/ route I added for metagov gets the api key header treatment also

express-server/src/types/context.ts
portiabytes: The original parsers weren't wrong per se, just using Zod's basic .url() validator, which only checked if something was a valid URL format but didn't enforce any of our security requirements as far as I could tell.

I would prefer to use a custom validator that does a few important things:
1. Uses the native URL constructor for more robust parsing
2. Enforces HTTPS in production (which the old parser wasn't doing)
3. Gives us more specific error messages that actually tell you which URL field failed
4. Makes the validation consistent across all our URL fields (CLIENT_BASE_URL, PYSERVER_URL, etc.)

The old parsers weren't catching cases where someone might try to use HTTP in prod, I think

express-server/src/types/context.ts
```diff
@@ -17,100 +17,155 @@ declare global {
...
+   */
+  requireHttps?: boolean;
+};
+
+/**
+ * Creates a URL validator with configurable HTTPS requirement
+ * @param fieldName - Name of the URL field for error messages
+ * @param config - Environment configuration options
+ */
+export const createUrlValidator = (fieldName: string, config: EnvConfig = {}) => {
```
staceysv, line 45: yes thank you!!!

express-server/src/types/context.ts
```diff
@@ -17,100 +17,155 @@ declare global {
...
+ */
+export const validateEnvironment = (env: string | undefined): Environment => {
+  if (!env || !["dev", "prod"].includes(env.toLowerCase())) {
+    throw new Error("Environment not set: NODE_ENV must be 'dev' or 'prod'");
+  }
+  return env.toLowerCase() as Environment;
+};
+
+/**
+ * Environment variable validation schema
```
staceysv, line 94: great, so much more clear!

staceysv:
@portiabytes this latest version looks great to me, thank you! ready to test locally if this is all the changes, staging instance may still take a little bit :)

README.md
```diff
@@ -166,6 +166,17 @@ Copy this file to .env.local in the same directory if you plan to run the dev se
 
 You can add different types of .env files based on your needs for testing, dev, prod, etc. You can read more in the [NextJS docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables)
 
+#### pyserver/.env
+
+Create a `.env` file in the `pyserver` directory with the following variables:
+
+```
+NODE_ENV=dev # Use 'dev', 'staging', or 'prod'
+OPENAI_API_KEY=your_openai_api_key
```
staceysv, line 175: add/note: any OpenAI API key you specify here will be used for your local development only / is only needed for "dev" mode
TODO: make sure this is traced/documented/correct in code flow :)

express-server/src/pipeline/claimsStep.ts
```diff
@@ -39,21 +40,24 @@ export async function claimsPipelineStep(
...
-        // TODO: how to set this in undici???
-        //redirect: "follow"
-        //signal: controller.signal,
-      });
+      };
+      
+      // Add redirect option for staging and production environments
+      if (requiresHttps(env.NODE_ENV as Environment)) {
+        // In undici, followRedirect is the equivalent of fetch's redirect: "follow"
+        requestOptions.followRedirect = true;
```
staceysv, line 57: awesome thank you!

pyserver/main.py
```diff
@@ -33,6 +33,10 @@
 from openai import OpenAI
 from pydantic import BaseModel
 import wandb
+import logging
+
+# Set up the logger
+logger = logging.getLogger(__name__)
```
staceysv, line 39: oh NICE hahhahahaa

pyserver/main.py
```diff
@@ -45,24 +49,32 @@
 
 class Environment(str, Enum):
     DEV = "dev"
+    STAGING = "staging"
```
staceysv, line 52: great! I'm still going back and forth on "staging" vs "stage"

next-client/src/lib/firebase/firestore.ts
```diff
@@ -14,15 +14,23 @@ import {
...
-const getCollectionName = useGetCollectionName(
-  NODE_ENV === "development" ? "dev" : "prod",
-);
+// Map NextJS environment values to T3C Environment type
+const mapNextEnvToT3CEnv = (nextEnv: string | undefined): Environment => {
+  if (!nextEnv) return "dev";
+  switch (nextEnv.toLowerCase()) {
+    case "development": return "dev";
+    case "production": return "prod";
+    case "test": return "dev";
```
staceysv, line 27: q: where do these long names, and also "test", get set in TypeScript? totally ok if we NEED to have two different sets of env names, and maybe we could have just one? :)

next-client/src/server-env.ts
```diff
@@ -84,3 +57,90 @@ if (!result.success) {
...
+
+export type ClientEnv = z.infer<typeof clientEnv>;
+
+/**
+ * Parse and validate client environment variables
+ */
+export function validateClientEnv(): ClientEnv {
+  const parsed = clientEnv.safeParse(process.env);
+
+  if (!parsed.success) {
```
staceysv, line 83: when I was working on this branch, I needed this change for types to be correct:
- if (!parsed.success) {
+ if (parsed.success === false) {

next-client/src/server-env.ts
```diff
@@ -84,3 +57,90 @@ if (!result.success) {
...
+ */
+export const clientEnv = z.object({
+  PIPELINE_EXPRESS_URL: z.string(),
+  NEXT_PUBLIC_FIREBASE_API_KEY: z.string(),
+  NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN: z.string(),
+  NEXT_PUBLIC_FIREBASE_PROJECT_ID: z.string(),
+  NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET: z.string(),
+  NEXT_PUBLIC_FIREBASE_MESSAGING_SENDER_ID: z.string(),
+  NEXT_PUBLIC_FIREBASE_APP_ID: z.string(),
+  NODE_ENV: z.string().optional().default("development"),
```
staceysv, line 72: why not "dev"?

express-server/src/types/context.ts
```diff
@@ -12,116 +17,21 @@ declare global {
...
+  FIREBASE_DATABASE_URL: z.string(),
+  FIREBASE_PROJECT_ID: z.string(),
+  CLIENT_BASE_URL: z.string(),
+  PYSERVER_URL: z.string(),
+  GOOGLE_CREDENTIALS_ENCODED: z.string().optional(),
+  GOOGLE_APPLICATION_CREDENTIALS: z.string().optional(),
+  REDIS_HOST: z.string(),
+  REDIS_PORT: z.string(),
+  REDIS_URL: z.string(),
+  NODE_ENV: z.string(),
```
staceysv, line 34: general note: are we sure about this list? at some point we need to describe each variable and where it comes form, and make sure we don't have any extras/unimplemented ones. Right now if I'm understanding correctly, we got rid of a bunch of validation...and just have one flat list?

if we don't have time to do that now/in this PR which is already hard to track, let's not make any changes unless they're already in scope & already necessary.
If we change any of these/set rules about them, let's do all of them solidly in this pass.

from my quick read, some pieces to resolve: 
- FIREBASE_PROJECT_ID: z.string() : this should be optional for now? doesn't get set via local env files/was crashing for me before on this check
- GOOGLE_APPLICATION_CREDENTIALS: what is this? seems like a duplicate of GOOGLE_CREDENTIALS_ENCODED, which definitely shouldn't be optional

staceysv:
;eft some comments, looking good, pausing review until further notice!
i'm increasingly wondering if we can split this into multiple PRs for our own reasonable management of all these changes :)

justinstimatze:
This work was split out and completed across #182 , #233, and #228. Closing this.

PR #100, opened 2025-02-09
Update About page with extended content 

colleenm:
Adding in the content Julia and Ethan have been working on for the about page, including FAQ, privacy stuff, and a brief report creation guide. (https://docs.google.com/document/d/192HHU9OYg4BqkWVIq1LRTIDJxGHdQqbuUcitJqVSLSk/edit?tab=t.pzrbyftahjmo)

@staceysv can you do a light content sanity pass if you have time?

@Brandon-Perry can you be primary reviewer on this?


colleenm:
PDF of about page for lazy review without running dev server

[Talk to the City about page.pdf](https://github.com/user-attachments/files/18726257/Talk.to.the.City.about.page.pdf)

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+        This alpha version of T3C accepts input of any unstructured text,
+        including free-form survey responses, interview and meeting transcripts,
+        and collections of social media posts. We have experimental features to
+        process audio and video datasets, and include video in reports &ndash;
+        reach out if you're interested in using these features while they're
+        under development.
+      </p>
+      <p>
+        For this alpha launch, all reports are <em>temporarily public</em> to
+        anyone who has the URL; we're adding password protection to reports
```
staceysv, line 116: would it be more or less helpful/reassuring to specify that the URL contains a unique SHA256 hash string which is virtually impossible to discover? maybe "For this alpha launch phase, any reports generated with T3C will be <em>publicly viewable</em> to anyone who has the (unique, SHA256-hashed) URL to the resulting report. We're adding password protection and allow-list permissions sharing to reports soon." ? Defer to you, I think this will be reassuring to a very niche groups of people — it's the same as a magic link/invitation, just they can't revoke it/we won't refresh it...

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+      <br />
+
+      <h5>What data format does the CSV file need to follow?</h5>
+      <ul className="list-disc list-inside pl-2">
+        <li>
+          Required column: "comment" (contains the text of the responses to
+          analyze with T3C)
+        </li>
+        <li>
+          Optional column: "interview" (respondent name, can be "Anonymous" or
```
staceysv, line 203: nit: it can't be Anonymous, as that would make ALL the rows attributed to the same speaker named "Anonymous" — maybe something like "(respondent name, can be a unique pseudonym per unique respondent, or leave blank for us to generate sequential anonymous ids for every row"
optionally add to clarify: any rows with the same "interview" entry will be attributed to the same speaker

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+        <li>
+          Optional column: “id” (for linking to any external data, will be
+          generated if blank)
+        </li>
+      </ul>
+      <p>
+        A{" "}
+        <a
+          className="underline"
+          href="https://docs.google.com/spreadsheets/d/15cKedZ-AYPWMJoVFJY6ge9jUEnx1Hu9MHnhQ_E_Z4FA/edit?gid=0#gid=0"
```
staceysv, line 215: nit: what's that magic you mentioned to open this in a new window? target="_blank" ?

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+            href="https://help.openai.com/en/articles/4936850-where-do-i-find-my-openai-api-key"
+          >
+            following the OpenAI instructions
+          </a>
+          .
+        </li>
+      </ul>
+      <p>
+        Note: API keys are never stored on T3C servers, and are sent in
+        SSL-secured headers to make sure they can't be intercepted through our
```
staceysv, line 251: nit: wondering if there's a higher-level/more reassuring explanation here that would still be accurate — "API keys are never stored on T3C servers and data access into and our of our app is SSL-encrypted" ?

next-client/src/app/about/page.tsx
colleenm: I think adding in acronyms people don't recognize makes the rest of the sentence land less well, and I really want "THING IS PUBLIC" to land, but I added the word "exact" here which gets at what you're saying

separately: I'm pretty convinced we want password protection, pretty unconvinced we want to add allow-list permissions. could maybe be convinced on the latter. if you feel differently I'll just say something more vague than "password protection" here

staceysv: exact url is perfect! agree with emphasis on public!!
I think the other thing I failed to capture is the "temporarily" makes it seem like.. this _particula_r url will not be public later... and I don't want people to think it'll be magically-made-not-public later unless they tell us/delete it/go in and change permissions later/etc.

password protection vs allow-list — no opinion, defer to you, especially if it's just for viewing and not editing/rerunning, password seems easier + makes sense.

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+        under development.
+      </p>
+      <p>
+        For this alpha launch, all reports are <em>temporarily public</em> to
+        anyone who has the URL; we're adding password protection to reports
+        soon.
+      </p>
+      <br />
+
+      <h3 id="case-studies">Case studies</h3>
```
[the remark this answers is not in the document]
colleenm: good point, added both links to written case studies and links to reports

next-client/src/app/about/page.tsx
colleenm: fixed

next-client/src/app/about/page.tsx
colleenm: changed

next-client/src/app/about/page.tsx
```diff
@@ -13,47 +13,399 @@ export default function AboutPage() {
...
+          generating your report.
+        </li>
+      </ul>
+      <br />
+
+      <h5>What we collect</h5>
+      <ul className="list-disc list-outside pl-6">
+        <li>
+          User input: text or queries entered into "Talk to the City" are
+          temporarily transmitted to OpenAI's API for generating responses. Raw
```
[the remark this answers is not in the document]
colleenm: fixed

colleenm:
pause review on this for now -- running into issues with local testing.

(if someone wants to pull this branch and test creating a report and it works, can be reviewed & merged)

Brandon-Perry (approved):

PR #102, opened 2025-02-10
changed barchart to show people instead of claims

Brandon-Perry:


colleenm (approved):
yesss looks good on the dev server, ship it! 🛥️

PR #109, opened 2025-02-15
Unified Test Case Infrastructure

portiabytes:
This PR introduces shared test cases between TypeScript and Python servers while (hopefully) maintaining their existing test implementations.

Note: It is not ready to merge, this is for visibility.

### 1. Shared Test Infrastructure
- Created standardized test case format in `common/test_cases.json`
- Moved hardcoded test data to shared location
- Structured test data to include:
  - Sample inputs
  - Expected outputs
  - Test cases for both topic trees and claims

### 2. Python Server Updates
- Updated `test_pipeline.py` to use shared test cases
- Added request validation tests
- Maintained mock system for basic testing
- Added support for optional API integration tests
- Added pytest configuration

### 3. Documentation
- Updated README to reflect test changes

## Testing Instructions
1. TypeScript tests:
```bash
cd express-server
npm test
```

2. Python tests:
```bash
cd pyserver
# With mocks
python -m pytest test_pipeline.py

# With API integration
export OPENAI_API_KEY=your-key
python -m pytest test_pipeline.py
```


pyserver/main.py
```diff
@@ -86,7 +89,7 @@ def read_root():
 ###################################
 # Step 1: Comments to Topic Tree  #
 #---------------------------------#
-@app.post("/topic_tree")
+@app.post("/topic_tree/")
```
staceysv, line 92: nit: we got rid of the trailing slashes because they cause a very annoying redirect in prod — could be missing something though!

pyserver/main.py
```diff
@@ -40,6 +40,9 @@
 
 app = FastAPI() 
 
+def get_client():
```
staceysv, line 43: smart, I like this :)

pyserver/test_pipeline.py
```diff
@@ -4,159 +4,105 @@
...
-  request ={"llm" : llm, "comments" : comments, "tree" : {"taxonomy" :tree}}
-  claims = client.post("/claims/?log_to_wandb=local_test_0", json=request).json()["data"]
-  json_print(claims)
-
-  print("\n\nStep 3: Dedup & sort\n\n")
-  llm.update({"user_prompt" : config.CLAIM_DEDUP_PROMPT})
-  request ={"llm" : llm, "tree" : claims }
-  full_tree = client.put("/sort_claims_tree/?log_to_wandb=local_test_0", json=request).json()["data"]
-  json_print(full_tree)
+# Remove hardcoded test data
```
staceysv, line 22: beautiful thank you!

pyserver/test_pipeline.py
```diff
@@ -4,159 +4,105 @@
...
-#test_wb_claims()
-#test_wb_dupes()
-#test_wb_full_pipeline(longer_pets_15)
 
-# TODO: test with edge case inputs
-# TODO: add assertions
-# assert response.status_code == 200
-# assert response.json() == {"name": "Foo", "price": 42}
+# Add Pydantic models for validation
+class Comment(BaseModel):
```
staceysv, line 35: nit: these are a subset of classes already defined in pyserver/main — maybe we import from there?

pyserver/test_pipeline.py
```diff
@@ -4,159 +4,105 @@
...
+    # Validate without making API call
+    try:
+        CommentsLLMConfig(**valid_request)
+        validation_passed = True
+    except ValidationError:
+        validation_passed = False
+    
+    assert validation_passed, "Valid request failed validation"
+
+@pytest.mark.skipif(not os.getenv("OPENAI_API_KEY"), reason="OpenAI API key not set")
```
staceysv, line 102: appreciate having this pattern, nice

pyserver/main.py
```diff
@@ -1170,6 +1163,15 @@ def cruxes_from_tree(req:CruxesLLMConfig, log_to_wandb:str = config.WANDB_GROUP_
 
   return {"data" : cruxes, "usage" : net_usage}
 
+@app.post("/generate_topic_tree")
```
staceysv, line 1166: nit: is this from testing?

staceysv:
Thanks so much, this is really helpful cleanup & organization!
Overall I'm very excited to have this way-more-neat Python setup and to unify it with TS. I think we'll need to add more cases to both/they may not always be identical, and I have some questions/uncertainty about how best to organize all the code — discussed below in categories of importance/priority, descending :)

**Running this:**
- worked for me with some python installations, maybe we can add those to the venv / have another go a the minimal setup/install steps, if any, to run the testing?
- I can't get this to work with an API key yet — were you able to? I get this warning, which isn't just the "skipped/not implemented" I expect?
```
stacey@Anastasias-MacBook-Pro pyserver % python -m pytest test_pipeline.py
======================================================================================= test session starts =======================================================================================
platform darwin -- Python 3.11.9, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/stacey/Code/aoi/t3c/tttc-light-js/pyserver
configfile: pytest.ini
plugins: anyio-4.8.0
collected 3 items                                                                                                                                                                                 

test_pipeline.py ..s                                                                                                                                                                        [100%]

======================================================================================== warnings summary =========================================================================================
../../../../../.pyenv/versions/t3c/lib/python3.11/site-packages/_pytest/config/__init__.py:1441
  /Users/stacey/.pyenv/versions/t3c/lib/python3.11/site-packages/_pytest/config/__init__.py:1441: PytestConfigWarning: Unknown config option: asyncio_mode
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

**More of a style/direction note/inspiration:** 
it would be very helpful to have an overview / summary / short list of the edge cases/types of issues we're already addressing / explicitly testing / what we want to add next — I do see this at several different levels, namely the function names and the comments inside pyserver/test_pipeline.py, and I'd love to see something like an outline/list in a docstring, or a readme, or printed out when the tests run? or even a long comment saying, e.g., we're testing input/output structure already, we've covered these 1-2 routes of the pyserver but not this other newer route, we want to test, e.g. jailbreaking attemtpts in CSV rows next, etc.

**Prioritized next steps:** 

P0
- common/src/types.ts seems useful/shorter and also overlaps with a bunch of existing zod-reliant definitions, like in common/schema.ts — let's unify these if we can/not create extra ones
- express-server/src/tests/pipeline.test.ts: I don’t think we should delete the older version, let's check with @Brandon-Perry how it's currently used & add new stuff there

P1
- common/src/apiPyserver.ts— why is this a separate file from the existing common/apiPyserver.ts? shall we combine these? The original is a bit more zod oriented, and I don't think we want to get rid of it—maybe we import these in different ways/different places?
- common/src/morphisms.ts and common/src/prompts.ts — why are these two separate files? can we put all of this in one prompts file? Also pretty sure Brandon was going to rename morphisms as a bit too inscrutable :)

P2
- test_cases.json in common is great — want to make sure/note somewhere that we’re careful about not expecting the same exact output if/when we call any LLMs
- changing lots of package versions/dependencies: I see a lot of small updates to various packages and versions — works on my machine, and so far only @Brandon-Perry has edited any of that—want to make sure this doesn't affect anything else :) / maybe we need to update some aspect of the build script...?
- express-server/src/config/environment.ts OOH this is very useful/probably what I need for a bunch of other configuration across dev/test/prod — I need to figure out where/how it gets used, also where is this .env.test file and what is in it?
- express-server/src/pipeline/index.ts: ahhh, interesting — so we’re calling the openai from typescript here directly, is this maybe for mocking...? i'm not sure what this is for or how it gets used — maybe it's unclear, we moved away from TS directly calling openai so we could iterate in Python / log to W&B / transparently so I could work on the pipeline without learning TS.

staceysv (commented):
This is an excellent cleanup pass and first PR — thanks for figuring out this sprawling repo!!! Left more detailed comments & questions, I think we'll need to do a bit more  careful merging with existing TS files / definitions & setup.

Brandon-Perry:
Can you explain more about the changes in `/common`? I don't quite understand what's being done. I also don't understand why `common/src/` exists.

justinstimatze:
Gently closing as recent and upcoming changes will cover some of the ideas here.

PR #111, opened 2025-02-18
Effect.TS / useReportState refactor

Brandon-Perry:
So I'm proposing that we introduce a new programming paradigm to our code based off of the library [Effect.TS.](https://effect.website/). It's a TypeScript based functional programming library. In this PR, I refactored the useReportState hook to give you all an idea as to what it would look like. 

Below I have a pros/cons list, along with a crash course in the parts of the library that I used. I would like to have a discussion about incrementally adopting this across our codebase. Also, I recommend checking out the branch and opening up `next-client/src/components/report/hooks/useReportState.ts` instead of looking at the git diff here. 

**Pros and cons**
Pros:
- _Cleaner code_:  It makes code that's easy to reason about. If you look at the original files, I had to write tests for a bunch of the custom functions that I wrote, I had to perform a bunch of existential checks, etc. I found that I was able to follow what I was writing a lot easier. 
- _Explicit error handling_: It makes it so we're not wrapping our code in a bunch of try/catch blocks. But instead we're handling our errors as explicit values. In my experience, that's wwwaaayyy better. 
- _Extensibility_: I think this will be a lot easy to build on top off. For example, if we wanted to add a size limit to CSVs, it's a lot easier to just add another line to the pipeline than what we currently do.
- _Less mistakes_: It enforces a paradigm that I think results in less foot-guns. I found with the previous code that it lead to lots of situations where I would code myself into a corner. I found that this happened less here.
- Incremental adoption: We can adopt it incrementally so we can make changes at our own pace. We also don't have to use it for every piece of code. 
- _Open source library_: The library is open source, so we can fork it if the group behind it dissolves.
- Our code is largely based around pipelines and transformations on data - so I think this is well suited for that.
- Functional programming is _based_

Cons:
- _Restrictions in talent pool_: FP is not as popular as OOP, so it limits our talent pool to a certain extent. Albeit since React went closer to FP a few years ago, it's a bit more common. I personally don't think it would be that big of a leap from standard React programming.
- _Difficulty_: This adds another level of difficulty for contributors. Although I think the mitigating factor here is that we can essentially choose how deep to go into the Effect library. Even if we stuck to the basics, I think we'd gain a lot.
- _Okay docs_: The docs could be better. But I've definitely seen worse. The main website docs cover most of what you need to know, but you have to do a little more digging to see the full extend of the APIs. I expect this will get better over time however. 

**Crash course in Effect API**j

There are a few main entities used in Effect: Option, Either, Array, Match, Effect, pipe, flow.

Option is essentially T | null.
```
const something = Option.some('something in here')
const missing = Option.none()

const head = <T>(arr:T[]):Option<T> => arr.length === 0 ? Option.none() : Option.some(arr[0]) 
``` 

Either is a way to manage errors as values. It uses Right | Left to differentiate the two.
```
const success = Either.right("it works!)
const myError = Either.left("D'oh!")

const safeDivide = (n,divisor) => divisor === 0 ? Either.left("Divide by 0 error") : Either.right(n/divisor)
```

Array is your usual JS Array class. Effect just has some extra fancy functions on it to make it more helpful.

Effect is for (generally) async code. It's essentially anything that is a side-effect in the code. It's not used in this PR because useReportState has no async code in it, so we don't need to go into detail. 

pipe is a way to create pipelines
```
const tenStr = pipe( // 10
   5,
   x => x * 2
   x => String(x)
)
```

flow is a way to do function composition
```
const double = x => x*2
const add5 = x => x + 5

const doubleAdd5 = flow(double, add5)
```

Match is just a fancy switch statement

X.map, where X is Either, Option, or Array, essentially "unpacks" the successful variant of the class to do something with it. Some examples:
```
pipe(
   Option.some(10),
   Option.map(x => x + 1)
} // Option.some(11)

pipe(
   Option.none(),
   Option.map(x => x+1)
} // Option.none

pipe(
   Either.right("My success"),
   Either.map(str => `It works!: ${str}`
) // Either.right("it works!: My success")

pipe(
   Either.left("Failed"),
    Either.map(str => `It works!: ${str}`
) // Either.left("Failed")
```

X.flatMap is like X.map, except that it "flattens" nested values, similar to Array.flatMap used in normal JS

```
Array.flatMap([
  [1],
  [2]
]) // [1,2]

Option.flatMap(
   Option.some(
      Option.some('x')
   )
) // Option.some('x')
It's okay if you don't really understand this one, it's used twice in this PR, and it's for arrays.
```
That should cover about 90% of what you need to know for this PR.


staceysv (approved):
Thanks so much, I like the direction! Ran the tests & read in detail. I support and I'm excited about this approach, and I'm still quite early in my TS journey, so this is mostly a "makes sense, let's try it!" endorsement. I think a more serious test of this setup will happen if/as we encounter & have to handle edge cases in createPathMapReducer and as we get more interesting action streams for fancier reports/new features we may not have anticipated (e.g., "open all claims in this report made by speaker=Ada").  

A few quick questions/notes:
- a bit confused by the name/existence of `makeSubSubtopicNode` instead of `makeSubtopicNode` — is this intentional/is there something that's more granular than a subtopic but not a claim?
- any thoughts on what would happen if we add more granularity to the tree eventually? E.g. another level of topic / theme below the current subtopic and above the current claim? seems like will be pretty straightforward / more straightforward in this world!
- similar consideration on using claim metadata (like duplicate claims, or speakers, or eventual additional metadata like which state the speaker lives in or something) and sorting by something different than nPeople — generally thinking about other ways we may want to sort/account for these types of details in TS and not in pyserver (e.g. I've found reports with 40+ claims in a subtopic are very hard to navigate without _some_ additional sorting within subtopic). No action for now, and hoping this is easier/faster/more flexible in this new world :)

Brandon-Perry:
> makeSubSubtopicNode

that's a typo, will fix before merging

> any thoughts on what would happen if we add more granularity to the tree eventually? 

I think that will be easy

> I think a more serious test of this setup will happen if/as we encounter & have to handle edge cases in createPathMapReducer and as we get more interesting action streams for fancier reports/new features we may not have anticipated (e.g., "open all claims in this report made by speaker=Ada").

That specific case at least would actually be very easy! Look closely at the MapAction type. It its a function ReportState -> ActionStreamActions[], so all you have to do is write a function that makes an open action for every node that has a certain condition, and dispatch it!

PR #114, opened 2025-02-21
Download JSON button

Brandon-Perry:
I added a button to the toolbar (temp placement) that downloads the report data for the user.

Waiting on designs for where the button should be placed before the PR is ready.

### DIFF
--- a/next-client/src/app/api/report/download/[uri]/route.ts
+++ b/next-client/src/app/api/report/download/[uri]/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from "next/server";
 
 export async function GET(_: Request, { params }: { params: { uri: string } }) {
-  const encodedUri = params.uri;
+  const encodedUri = await params.uri; // we get an error saying that params needs to be awaited? Add await here until it's clearer as to what's going on.
   const uri = decodeURIComponent(encodedUri);
   const jsonData = await fetch(uri, {
     method: "GET",
--- a/next-client/src/app/api/report/download/[uri]/route.ts
+++ b/next-client/src/app/api/report/download/[uri]/route.ts
@@ -1,0 +1,14 @@
+import { NextResponse } from "next/server";
+
+export async function GET(_: Request, { params }: { params: { uri: string } }) {
+  const encodedUri = params.uri;
+  const uri = decodeURIComponent(encodedUri);
+  const jsonData = await fetch(uri, {
+    method: "GET",
+    headers: {
+      "Content-Type": "application/json",
+    },
+  }).then(async (res) => await res.json());
+
+  return NextResponse.json(jsonData);
+}
--- a/next-client/src/app/api/report/download/[uri]/route.ts
+++ b/next-client/src/app/api/report/download/[uri]/route.ts
@@ -1,4 +1,28 @@
 import { NextResponse } from "next/server";
+import { getReportDataObj } from "tttc-common/morphisms/pipeline";
+import * as schema from "tttc-common/schema";
+
+type ReportData = ["report", schema.UIReportData];
+type Error = ["error"];
+
+const handleParsingJson = (data: unknown): ReportData | Error => {
+  if (schema.llmPipelineOutput.safeParse(data).success) {
+    // if the data is from the old schema, then translate it into the new one
+    const newSchemaData = getReportDataObj(
+      schema.llmPipelineOutput.parse(data),
+    );
+    return ["report", schema.uiReportData.parse(newSchemaData)];
+  } else if (schema.pipelineOutput.safeParse(data).success) {
+    return [
+      "report",
+      schema.uiReportData.parse(schema.pipelineOutput.parse(data).data[1]),
+    ];
+  } else if (schema.downloadReportSchema.safeParse(data)) {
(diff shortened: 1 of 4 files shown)


next-client/src/app/api/report/download/[uri]/route.ts
```diff
@@ -0,0 +1,14 @@
+import { NextResponse } from "next/server";
+
+export async function GET(_: Request, { params }: { params: { uri: string } }) {
```
Comment (staceysv) [inline next-client/src/app/api/report/download/[uri]/route.ts:3]: I see this error when running locally — doesn't hurt performance, and maybe we should fix it?
`Error: Route "/api/report/download/[uri]" used `params.uri`. `params` should be awaited before using its properties. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
    at uri (src/app/api/report/download/[uri]/route.ts:4:28)
  2 |
  3 | export async function GET(_: Request, { params }: { params: { uri: string } }) {
> 4 |   const encodedUri = params.uri;
    |                            ^
  5 |   const uri = decodeURIComponent(encodedUri);
  6 |   const jsonData = await fetch(uri, {
  7 |     method: "GET",
`

next-client/src/components/report/Report.tsx
```diff
@@ -164,6 +171,7 @@ function Report({ reportData }: { reportData: schema.ReportDataObj }) {
               {state.children.map((themeNode) => (
                 <Theme key={themeNode.data.id} node={themeNode} />
               ))}
+              <Appendix filename={reportData.title} reportUri={reportUri} />
```
staceysv, line 174: nit: super minor/can work around this, and:
- we don't enforce unique titles for reports
- this may cause the download to overwrite if people are trying it for the same test report/multiple-times-quickly

could we append a quick timestamp into the file name?

staceysv (changes requested):
awesome, thank you so much for the fast turnaround! this works for me now & I think is generally great, one blocking thing & some fast follows/next steps at lower priority

Blocking this PR
- let's redact the "sources" field from the json

High-priority / fast-follow / to discuss
- @colleenm suggests we may want to make this "download json" button only visible to report creators — I don't have a super strong opinion because so few reports are findable without them being effectively public. How hard would it be to only show the button on 
A. A user's own reports they created or 
B. also any public "example demo" reports like our 5 case studies on the landing page?

Notes / non-blocking / fast-follows
- got an error on the route, not breaking and surprising?
- let's leave out the "metadata" field entirely until/unless we can fill it correctly.
- would be nice to get a unique name for the file, add timestamp to report title?
- can we change the "date" value to be the "date created" of the report, or if that's too hard, make the key name "date_downloaded" to avoid confusion?
- could we get a quick writeup of the format for this file in a readme in the repo —examples/README.md ideally?

 I see the following, feel free to copy & paste as a start, e.g.

The resulting json is a dictionary with two primary keys:
- "data": a list of dictionaries, where the key is the version (like "v0.2") and the value is a dictionary, containing "title", "description", "topics": [ full report ], "sources" : [ list of raw report data], "date" : (of DOWNLOAD, not of report being made!!)
- "metadata" : as follows, currently empty
    "metadata": [
        "v0.2",
        {
            "duration": 0,
            "startTimestamp": 1740526265355,
            "totalCost": "",
            "author": "",
            "organization": ""
        }
    ]

next-client/src/app/api/report/download/[uri]/route.ts
Comment (Brandon-Perry) [reply next-client/src/app/api/report/download/[uri]/route.ts:3]: I added an await there. I don't remember getting an error what that happened. Can you check to make sure that that solves the error on your end?

next-client/src/components/report/Report.tsx
Brandon-Perry: Added

next-client/src/app/api/report/download/[uri]/route.ts
Comment (staceysv) [reply next-client/src/app/api/report/download/[uri]/route.ts:3]: thank you, will check!

staceysv (approved):
awesome!!

next-client/src/app/api/report/download/[uri]/route.ts
Comment (staceysv) [reply next-client/src/app/api/report/download/[uri]/route.ts:27]: hmmm this is still giving an error, but the download succeeds! making a note to investigate later
```
✓ Compiled /api/report/download/[uri] in 137ms (1480 modules)
Error: Route "/api/report/download/[uri]" used `params.uri`. `params` should be awaited before using its properties. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis
    at uri (src/app/api/report/download/[uri]/route.ts:28:34)
  26 |
  27 | export async function GET(_: Request, { params }: { params: { uri: string } }) {
> 28 |   const encodedUri = await params.uri; // we get an error saying that params needs to be awaited? Add await here until it's clearer as to what's going on.
     |                                  ^
  29 |   const uri = decodeURIComponent(encodedUri);
  30 |   const jsonData = await fetch(uri, {
  31 |     method: "GET",
 GET /api/report/download/https%3A%2F%2Fstorage.googleapis.com%2Ftttc-test-2-bucket%2Fb485442fbedaaa61206fd6caf0167921f72f489c3c9c9bf984c799ceb2ee7387 200 in 537ms
```

PR #122, opened 2025-03-07
compute and log cost and tokens for LLM pipeline

staceysv:
TLDR: Sum and log the costs and token counts for each step of the LLM pipeline

We already have a "tracker" in workers.ts — I added a quick function to calculate the cost of each pipeline step and actually add that cost, and the associated token counts, to the tracker.
My goal was to implement this quickly — I very much welcome refactoring/cleanup for TS style (ideally as a PR on top of this cc @Brandon-Perry )

* calculate costs in pyserver, using a quick dictionary of prices per model and a utils function for the math (returns -1 if we use a model for which we don't yet store the price)
* return the cost of each LLM pipeline step from pyserver (and log to wandb when relevant)
* return the cost from the TS wrapper for each pipeline step
* in workers.ts, add a helper function that accumulates the total cost and token counts for each report generation, and log the new totals whenever it's called

### DIFF
--- a/common/apiPyserver.ts
+++ b/common/apiPyserver.ts
@@ -171,6 +171,7 @@ export const topicTreeResponse = z.object(
   {
     data: partialTopic.array(),
     usage,
+    cost: z.number(),
   },
   { invalid_type_error: "Invalid topic tree response" },
 );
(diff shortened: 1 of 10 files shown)


staceysv:
Thanks @Brandon-Perry !

Brandon-Perry (approved):

PR #124, opened 2025-03-11
correctly fix timeout for long reports

staceysv:
creating a record of @objectivevan 's fix to enable long reports in production!

todos before merges (mostly for @Brandon-Perry 's TS wisdom!)

- [ ] please add any TS style changes 
- [ ] reduce the console logs to only the ones indicating errors/error messages
- [ ] are all the package.json changes expected/needed/correct?

### DIFF
--- a/express-server/src/pipeline/claimsStep.ts
+++ b/express-server/src/pipeline/claimsStep.ts
@@ -47,7 +47,7 @@ export async function claimsPipelineStep(env: Env, input: ClaimsStep["data"]) {
         //signal: controller.signal,
       });
       console.log("POST call finished");
-      // Clear the custom timeout since we've received a response
+
       clearTimeout(timeoutId);
 
       // Check status code (similar to `response.ok`)
@@ -84,7 +84,7 @@ export async function claimsPipelineStep(env: Env, input: ClaimsStep["data"]) {
     }
   } catch (error) {
     console.error("Claims pipeline step failed:", error);
-    // Re-throw or return an error response as desired
+
     throw error;
   }
 }
--- a/express-server/src/pipeline/claimsStep.ts
+++ b/express-server/src/pipeline/claimsStep.ts
@@ -2,38 +2,89 @@ import * as apiPyserver from "tttc-common/apiPyserver";
 import { ClaimsStep } from "./types";
 import { Env } from "../types/context";
 import { z } from "zod";
-
-const typedFetch =
-  <T extends z.ZodTypeAny>(bodySchema: T) =>
-  async (url: string, body: z.infer<T>) =>
-    await fetch(url, {
-      method: "POST",
-      body: JSON.stringify(bodySchema.parse(body) as z.infer<T>),
-      headers: {
-        "Content-Type": "application/json",
-      },
-      // wait for 7 minutes for full claims list
-      // TODO: use message queue instead
-      signal: AbortSignal.timeout(1200000),
-    });
-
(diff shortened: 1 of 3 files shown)


express-server/src/pipeline/claimsStep.ts
```diff
@@ -2,38 +2,89 @@ import * as apiPyserver from "tttc-common/apiPyserver";
 import { ClaimsStep } from "./types";
 import { Env } from "../types/context";
 import { z } from "zod";
+import { Agent } from "undici";
```
Brandon-Perry, line 5: Agent and jsonrepair are not used so these should be removed

package.json
Brandon-Perry, line 1: These changes shouldn't be in root, it should be in express-pipeline

express-server/src/pipeline/claimsStep.ts
```diff
@@ -2,38 +2,89 @@ import * as apiPyserver from "tttc-common/apiPyserver";
...
-const logger =
-  (prependMessage: string) =>
-  <T>(arg: T): T => {
-    console.log(prependMessage, arg);
-    return arg;
-  };
+      // Check status code (similar to `response.ok`)
+      if (statusCode < 200 || statusCode >= 300) {
+        // Read error text from the response body
+        let errorText = "";
```
Brandon-Perry, line 56: I'd use `await body.json()` instead of constructing the response string piecemeal unless there's a good reason to do otherwise.

express-server/src/pipeline/claimsStep.ts
```diff
@@ -2,38 +2,89 @@ import * as apiPyserver from "tttc-common/apiPyserver";
...
-    `${env.PYSERVER_URL}/claims`,
-    input,
-  )
-    .then((res) => res.json())
-    .then(logger("claims step returns: "))
-    .then(apiPyserver.claimsReply.parse);
-
-  return { claims_tree: data, usage };
+      // Read the response body
+      let responseData = "";
```
Brandon-Perry, line 64: Same as the above comment about using body.json()

express-server/src/workers.ts
```diff
@@ -217,7 +219,7 @@ const setupPipelineWorker = (connection: Redis) => {
         status: api.reportJobStatus.Values.finished,
       });
     },
-    { connection },
+    { connection, stalledInterval: 3000000, skipStalledCheck: true },
```
Brandon-Perry, line 222: Can you add a comment explaining the decision around this? Also wouldn't skipStalledCheck mean we don't need to set a stalledInterval?

package.json
```diff
@@ -1,19 +1,49 @@
...
+    "dotenv": "^16.4.5",
+    "express": "^4.18.3",
+    "firebase-admin": "^13.0.2",
+    "ioredis": "^5.4.1",
+    "node-fetch": "^3.3.2",
+    "openai": "^4.28.4",
+    "prettier": "^3.2.5",
+    "tttc-common": "file:../common/dist",
+    "undici": "^7.0.0",
+    "jsonrepair": "^1.0.0",
```
Brandon-Perry, line 45: I don't think jsonrepair is being used anywhere? Which case it should be removed.

Brandon-Perry (changes requested):

express-server/src/workers.ts
staceysv: official answer: this is the only combination that works empirically!

Brandon-Perry (approved):

PR #126, opened 2025-03-13
About page fixes

Brandon-Perry:


staceysv:
Tested locally, style updates look good to me!
I'm wondering if this is the right time to iterate on content / collapse some of it — also ok just merging these fixes.

staceysv (approved):
lgtm! could iterate more on content now or later

PR #130, opened 2025-03-18
[WIP/DNM] crux extraction is optional from UI

staceysv:
- make the crux extraction step optional
- match crux prompting logic to other pipeline steps (so users can see & override the prompt)

- moved crux logic to match other prompt instructions
- tried to add a checkbox  — ideally we'd check the value of this checkbox and only run the crux extraction step if a user actively checked that box

- working so far on localhost
- need to figure out how to correctly set the default value to false and pass it through to workers.ts if the user checks the box

### DIFF
--- a/common/prompts.ts
+++ b/common/prompts.ts
@@ -81,6 +81,23 @@ And now, here are the claims:
 \${claims}
 `;
 
+export const defaultCruxPrompt = `
+I'm going to give you a topic with a description and a list of high-level claims about this topic made by different participants,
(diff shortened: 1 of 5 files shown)


staceysv:
> I think it would be good practice to set this as a draft PR if it's not ready for merging.

@Brandon-Perry thanks good reminder, thought I did!

Brandon-Perry:
@staceysv also I notice there's a a bunch of linting commits. Try running `npm run format` at the root level before adding and committing. If you forget, you can add again and run `git commit --amend --no-edit` if you haven't pushed, and it'll retroactively add those changes.

PR #135, opened 2025-03-24
Express server strict mode

Brandon-Perry:
Make it so we're using strict:true on the express-server, which makes type-checking stronger 
Changed ts-config in express/common to use strict:true, changed places in code that started to throw type errors.

Note: in some places I used ts-ignore, which is generally not good practice, but those places desperately need refactor / would have been a headache to do. 
Ran pets

### DIFF
--- a/express-server/src/googlesheet.ts
+++ b/express-server/src/googlesheet.ts
@@ -16,7 +16,8 @@ export async function fetchSpreadsheetData(
   // extract the data from the spreadsheet
   const url2 = `https://docs.google.com/spreadsheets/d/${spreadsheetId}/gviz/tq?tqx=out:json`;
   const response = await axios.get(url2);
-  const jsonStr = response.data.match(/(?<=.*\().*(?=\);)/s)[0];
+  const match = response.data.match(/.*\(([\s\S]*)\);/);
+  const jsonStr = match ? match[1] : null;
   const json = JSON.parse(jsonStr);
 
   // extract the columns and rows
@@ -63,7 +64,11 @@ export async function fetchSpreadsheetData(
   const emailToData: { [key: number]: { id: string; comment: string }[] } = {};
 
   rows.forEach((row, id) => {
+    // ! Brandon: indexing into a string and then using that to access a record with key:number ?????????
+    // ! Brandon: Adding ts ignores into this so we can use strict mode and not break stuff. This entire file needs redone.
+    // @ts-ignore
     emailToData[row[emailColumn]] ??= [];
+    // @ts-ignore
     emailToData[row[emailColumn]].push({
       id: String(id),
       comment: commentColumns
--- a/express-server/src/pipeline.ts
+++ b/express-server/src/pipeline.ts
@@ -37,6 +37,7 @@ function insertClaim(taxonomy: Taxonomy, claim: LLMClaim, tracker: Tracker) {
   const matchedTopic = taxonomy.find((topic) => topic.topicName === topicName);
   if (!matchedTopic) {
     console.log("Topic missmatch, skipping claim " + claim.claimId);
+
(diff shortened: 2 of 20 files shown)


express-server/src/googlesheet.ts
```diff
@@ -63,7 +64,11 @@ export async function fetchSpreadsheetData(
   const emailToData: { [key: number]: { id: string; comment: string }[] } = {};
 
   rows.forEach((row, id) => {
+    // ! Brandon: indexing into a string and then using that to access a record with key:number wtf?!?
+    // ! Brandon: Adding ts ignores into this so we can use strict mode and not break shit. This entire file needs redone.
```
staceysv, line 68: really appreciate the comment & understand the feeling!
nit/general reminder, our repo is open source, let's keep the language more professional
e.g. "access a record with key:number — TODO: investigate why/how"
e.g. "so we can use strict mode and not break anything — TODO: redo this entire file" (added to task tracking btw)

express-server/src/pipeline.ts
```diff
@@ -192,7 +194,7 @@ async function pipeline(
       : `${secs} seconds`;
 
   // next line is important to avoid leaking keys!
-  delete options.apiKey;
+  options.apiKey = "";
```
staceysv, line 197: general: are we using this file at all? I thought we're only using workers.ts for the pipeline these days? can we get rid of this whole file if yes?

express-server/src/routes/report.ts
```diff
@@ -7,6 +7,10 @@ export async function report(req: Request, res: Response) {
     api.getReportRequestUri.parse(req.params.reportUri),
   );
   const name = new URL(uri).pathname.split("/").pop();
+  if (!name) {
```
staceysv, line 10: can we add a comment on what this state is/what this means? and/or log what this means — malformed request or bad input url / unable to extract report name?

staceysv (changes requested):
Looks good, thanks for this cleanup! Left a few minor comments in places I would like a bit more explanation or see an opportunity for more cleanup (e.g. deleting deprecated files).
Q1: shall we test this a bit more before merging to prod? Any edge cases you can anticipate?
Q2: Also, how temporary do you think these typescript workarounds are/how urgent is it to refactor?

express-server/src/pipeline.ts
Brandon-Perry: will remove in a future PR

express-server/src/routes/generate.ts
```diff
@@ -16,13 +16,20 @@ export default async function generate(req: Request, res: Response) {
   try {
     const body: GenerateApiRequest = req.body;
     const { data: dataPayload, userConfig } = body;
```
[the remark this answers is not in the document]
Brandon-Perry: Will remove in a future PR

Brandon-Perry:
As far as testing goes - I think it's fine. There are no real changes to the logic, it's all type changes. If there as an issue with it, it would have thrown while transpiling. 

So I do think the backend needs a pretty extensive refactor. The issue with ts-ignore (which is used a lot in this PR) is that we lose the benefit of TypeScript in that spot. So if, for example, we change a type, it could mean that there's an error there that we're not catching. The issue however, is that the way that the backend is currently written would mean that more extensive changes would need to be made.
In terms of priority, I definitely say it should be a priority. But, that doesn't necessarily mean that we drop everything and rewrite it. I think doing so in pieces is more maintainable. So as we're implementing new features (like the private / public buckets), we should be refactoring as we go, instead of just trying to jerry-rig it.

staceysv:
Beautiful, thanks for all the responses & comments :) — let's make sure to keep refactoring as we go!

staceysv (approved):

PR #138, opened 2025-03-26
Private gbuckets

Brandon-Perry:
This adds the ability to run both public and private google buckets. However, it does not implement the ability to add or read from the private bucket yet. 
Refactored storage.ts to use an abstract storage class and a bucket class
Ran a report manually.

### DIFF
--- a/express-server/src/storage.ts
+++ b/express-server/src/storage.ts
@@ -6,6 +6,7 @@ import { CustomError } from "./error";
 import * as schema from "tttc-common/schema";
 import { z } from "zod";
 import { Result } from "./types/result";
+import { Env } from "./types/context";
 
 const fileContent = z.union([schema.pipelineOutput, schema.uiReportData]);
 type FileContent = z.infer<typeof fileContent>;
@@ -75,12 +76,14 @@ export class Bucket extends Storage {
   async save(
     fileName: string,
     fileContent: string,
-  ): Promise<Result<string, BucketSaveError>> {
+  ): Promise<Result<string, StorageSaveError>> {
     const file = this.bucket.file(fileName);
     try {
       await file.save(fileContent, {
         metadata: {
           contentType: "application/json",
+          // Cache control here is required to successfully save to bucket for some reason.
+          cacheControl: "no-cache, no-store, must-revalidate",
         },
       });
       return {
@@ -117,3 +120,8 @@ class BucketSaveError extends CustomError<"BucketSaveError"> {
     super("BucketSaveError", err);
   }
 }
+
+export const createStorage = (env: Env): Storage => {
+  // since Bucket is the only storage class, just return this for now.
+  return new Bucket(env.GOOGLE_CREDENTIALS_ENCODED, env.GCLOUD_STORAGE_BUCKET);
+};
--- a/express-server/src/storage.ts
+++ b/express-server/src/storage.ts
@@ -1,39 +1,100 @@
-import { Storage } from "@google-cloud/storage";
+import {
+  Storage as BucketStorage,
+  Bucket as GBucket,
+} from "@google-cloud/storage";
+import { CustomError } from "./error";
+import * as schema from "tttc-common/schema";
+import { z } from "zod";
 
-let storage: any;
-const encoded_creds = process.env.GOOGLE_CREDENTIALS_ENCODED;
-if (encoded_creds) {
-  const decoded = Buffer.from(encoded_creds, "base64").toString("utf-8");
-  storage = new Storage({ credentials: JSON.parse(decoded) });
+const fileContent = z.union([schema.pipelineOutput, schema.uiReportData]);
+type FileContent = z.infer<typeof fileContent>;
+
+/**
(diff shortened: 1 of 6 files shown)


express-server/src/storage.ts
```diff
@@ -1,39 +1,139 @@
...
+
+export const createStorage = (
+  env: Env,
+  control: "public" | "private",
+): Storage => {
+  // since Bucket is the only storage class, just return this for now.
+  if (control === "public")
+    return new Bucket(
+      env.GOOGLE_CREDENTIALS_ENCODED,
+      env.GCLOUD_STORAGE_BUCKET,
```
staceysv, line 132: minor: could be more clear if we called this GCLOUD_STORAGE_BUCKET_PUBLIC to distinguish from GCLOUD_STORAGE_BUCKET_PRIVATE. Though this would require updating all env files to use the _PUBLIC env var instead.

express-server/src/storage.ts
```diff
@@ -1,39 +1,139 @@
...
+    );
+  else
+    return new Bucket(
+      env.GOOGLE_CREDENTIALS_ENCODED,
+      env.GLCOUD_STORAGE_BUCKET_PRIVATE,
```
staceysv, line 137: typo: should be GCLOUD

express-server/src/types/context.ts
```diff
@@ -30,6 +30,9 @@ export const env = z.object({
   GCLOUD_STORAGE_BUCKET: z.string({
     required_error: "Missing GCloud storage bucket",
   }),
+  GLCOUD_STORAGE_BUCKET_PRIVATE: z.string({
```
staceysv, line 33: typo: should be GLCOUD

express-server/src/storage.ts
```diff
@@ -1,39 +1,139 @@
...
+    fileName: string,
+    fileContent: string,
+  ): Promise<Result<string, StorageSaveError>> {
+    const file = this.bucket.file(fileName);
+    try {
+      await file.save(fileContent, {
+        metadata: {
+          contentType: "application/json",
+          // Cache control here is required to successfully save to bucket for some reason.
+          cacheControl: "no-cache, no-store, must-revalidate",
```
staceysv, line 86: question: it looks like in the earlier version of the code (workers.ts:286), we were setting allowCache to true, which would have skipped this cache control setting and cached instead. Do we know what changed?

staceysv (changes requested):
looks good! left minor comments that need to be addressed before we merge (typo + cache explanation)

express-server/src/storage.ts
Brandon-Perry: So previously how it worked is that storeJSON gets called twice: to store the initial "we're working on your report" message into the file, and then inside workers to store the actual report. The first time it's called, it doesn't have this "allowCache" param, so it uses this the cache control that we have defined above. The second time it allows for caching. This is because if we allowed caching on the first call, users would get shown stale data until the cache expires.

I omitted caching entirely because if we're going to allow for editing our JSON data, we can't have caching, or else we run the risk of showing stale data. We could loosen the cache controls to remove 'no-store' which prevents caching of any kind, so we can use caching with re-validation. However, I want more robust tests or something that increases my confidence that it won't break before implementing that. You can [read more about cache controls here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#directives).

"There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors."

staceysv: ah, thank you for explaining, esp the part about the two different modes for the phases of "tell the user we're working on the report" vs "actually save the report".

re: editing our JSON data, agree we don't want to show stale data — this current approach without caching seems right to me for now. Perhaps we could revisit eventually if we want to show, e.g, large detailed intermediate states of report generation to the user and cache those...? cc @justinstimatze

staceysv:
thanks @Brandon-Perry, looks good to merge to me!
One important note is that when we deploy this, we'll have to carefully rename the env variable GCLOUD_STORAGE_BUCKET to GCLOUD_STORAGE_BUCKET_PUBLIC throughout manual/automated deploys & READMEs/local setup notes going forward.

staceysv (approved):

Brandon-Perry:
@staceysv Lets add those README changes to this branch then? I don't know how infra stuff would work off the top of my head.

staceysv:
@Brandon-Perry let's hold off on the readme changes/the env var renaming for now, this enables us to merge to main / not worry about changing deploy process at all

express-server/src/storage.ts
justinstimatze: Apologies for the back and forth, I believe we decided to keep this as GCLOUD_STORAGE_BUCKET and rename it in a future PR once env changes are simpler.

justinstimatze (changes requested):
One small change for the env var name, otherwise LGTM.

Brandon-Perry:
Reverted the change

justinstimatze (approved):
LGTM

PR #139, opened 2025-03-31
fixed husky hook so that it runs before committing

Brandon-Perry:
Fixes the issue where the precommit hook was running after completing the commit
just the husky sh file
Ran a bunch of test commits with both files that get formatted and those that don't

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,21 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run format
+# Get all staged files that are JavaScript, TypeScript, JSX, TSX, CSS, SCSS, HTML, JSON, or MD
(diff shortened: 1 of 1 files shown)


staceysv (changes requested):
this is the right direction — let's make sure someone can reproduce this as a new dev! 
1. Probably I need to install husky so it runs from the command line?
2. Are there other tools like this/already a list of things to do for dev setup for npm / TS to work as expected from command-line?
3. Can we start to make this list of instructions and put it in the README?

Brandon-Perry:
@staceysv I don't know what you mean about installing husky. I checked this out in a clean repo, ran `npm i`, and everything worked.

staceysv:
tested this a bunch — the gotcha is to run `npm i` from root, in addition to/beyond running `npm i && npm run build` from `common/` as you normally would for local dev.
in the future, if we could see a list of the changes that prettier makes, that would be cool!

staceysv (approved):

PR #141, opened 2025-03-31
minimal setup for cruxes checkbox

staceysv:
- Add a checkbox in createReport which controls if we run the crux step
- Only run the crux step if the user checks the box
- Let users customize the crux step prompt in advanced settings 

- Input form: adds a checkbox
- Makes the crux step of the pipeline optional

- tested and seems to work locally (could use a bit more testing once we fix/mock LLM calls)

Most interested in any more graceful ways to:
- let zod accept a boolean that can be null: I ended up with the following, maybe this could be simpler? `z.boolean().optional().nullish().transform((val) => val ?? false)`
- define cruxesEnabled in one place in the code instead of three as I am currently doing — maybe only one is necessary/this is code we can clean up?
- run the optional step in workers.ts without defining dummy variables

### DIFF
--- a/common/schema.ts
+++ b/common/schema.ts
@@ -71,6 +71,7 @@ export const llmUserConfig = z.object({
   extractionInstructions: z.string().min(1),
   dedupInstructions: z.string().min(1),
   cruxInstructions: z.string().optional(),
+  cruxesEnabled: z.boolean().optional(),
 });
 
 export type LLMUserConfig = z.infer<typeof llmUserConfig>;
@@ -96,6 +97,7 @@ export const oldOptions = z.object({
   extractionInstructions: z.string(),
   dedupInstructions: z.string(),
   cruxInstructions: z.string().optional(),
+  cruxesEnabled: z.boolean().optional(),
   batchSize: z.number(),
   filename: z.string(),
   googleSheet: z
--- a/common/schema.ts
+++ b/common/schema.ts
@@ -71,7 +71,7 @@ export const llmUserConfig = z.object({
   extractionInstructions: z.string().min(1),
   dedupInstructions: z.string().min(1),
   cruxInstructions: z.string().optional(),
-  cruxesEnabled: z.boolean().optional(),
+  cruxesEnabled: z.boolean().optional().nullish().transform((val) => val ?? false)
 });
 
 export type LLMUserConfig = z.infer<typeof llmUserConfig>;
@@ -97,7 +97,7 @@ export const oldOptions = z.object({
   extractionInstructions: z.string(),
   dedupInstructions: z.string(),
   cruxInstructions: z.string().optional(),
-  cruxesEnabled: z.boolean().optional(),
+  cruxesEnabled: z.boolean().optional().nullish().transform((val) => val ?? false),
   batchSize: z.number(),
   filename: z.string(),
   googleSheet: z
--- a/express-server/src/workers.ts
+++ b/express-server/src/workers.ts
@@ -171,6 +171,7 @@ const setupPipelineWorker = (connection: Redis) => {
       // us refer to the same tracker objects — open to more graceful solutions and
       // this is why I wanted to pass by reference ~S
       const tracker_crux = tracker_step2;
+      let cruxAddOns = {};
       if (options.cruxesEnabled === true) {
 
         console.log("Step 2.5: Optionally extract cruxes");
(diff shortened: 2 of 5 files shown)


staceysv:
fixed the default value! seems to work now!
would like to test more/open to cleanup feedback :)

common/schema.ts
Brandon-Perry, line 1: Is there a reason why we should expect/want cruxes enabled to be either undefined or null? This should probably just be a boolean.

express-server/src/workers.ts
```diff
@@ -163,34 +164,42 @@ const setupPipelineWorker = (connection: Redis) => {
...
-        stepUsage: cruxTokens,
-        stepCost: cruxCost,
-      });
-      logTokensInTracker(tracker_crux);
-
+      console.log("user enabled crux extraction: ", options.cruxesEnabled);
+
+      // TODO: more graceful way to catch the case where we don't run Step 2.5?
+      // perhaps we can pass by reference instead of renaming the tracker object?
+      let cruxAddOns = {};
```
Brandon-Perry, line 171: I would caution against mutating variables like this. Instead, I'd write a function something like:
`function CruxStep(cruxProps, tracker):Promise<{CruxAddons, Tracker}> {...}`

Or something like that.

express-server/src/workers.ts
```diff
@@ -163,34 +164,42 @@ const setupPipelineWorker = (connection: Redis) => {
...
+          top_k: 0,
+        });
+        // package crux addOns together
+        const cruxAddOns = {
+          topCruxes: topCruxes,
+          controversyMatrix: controversyMatrix,
+          cruxClaims: cruxClaims,
+        };
+
+        const tracker_crux = sumTokensCost({
```
Brandon-Perry, line 195: I'm don't think this would work. These constants are declared in a different lexical scopes, so the inner is not overwriting the outside one.

next-client/src/components/create/CreateReport.tsx
```diff
@@ -86,6 +87,8 @@ const form = z.object({
   clusteringInstructions: z.string().min(1),
   extractionInstructions: z.string().min(1),
   dedupInstructions: z.string().min(1),
+  cruxInstructions: z.string().optional(),
```
Brandon-Perry, line 90: I don't think the crux instructions should be optional. The issue is that it introduces a state that can cause a failure. If crux enabled is set to true, but the instructions are undefined, then the pipeline fails. If make it so the instructions are required (even if not used), then we reduce the number of edge cases we need to lookout for.

Brandon-Perry: Also what do you think about adding a .min(1) to it? I remember you said something about the other ones causing an issue?

Brandon-Perry (changes requested):

common/schema.ts
staceysv: good point! I tried to just do boolean, and on first load it would often show up as empty/undefined — maybe boolean with default(false) would be the right move?

Brandon-Perry: Checkout SubmitAction.ts and see how it's making the API request to the express server. Odds are it's just missing on that. You'll probably need to add the field to the type in common/api

staceysv: hmmm, ok, I think it's just this: cruxesEnabled: formData.get("cruxesEnabled"),
and I made it non-optional but kept the nullish/make a null false:
z.boolean().default(false).nullish().transform((val) => val ?? false)

from my reading, this can happen with checkboxes that default to null/not set until they ARE clicked in

express-server/src/workers.ts
staceysv: sorry, you mean, add a new function, pass it the tracker, and return cruxAddOns and tracker from it?
that seems fine / maybe closer to what you're doing in the refactor already? we just need to call that step conditionally.

express-server/src/workers.ts
staceysv: oh, so should this one on 195 be let tracker_crux? so that we can overwrite it in the inner one?
I'm just trying to get this to work given that we're currently renaming the tracker variable after every step, which is not how I would do it.

next-client/src/components/create/CreateReport.tsx
staceysv: i remember you saying we wanted all crux stuff optional because it's new/only happening in some reports and we don't want to worry about backwards compatibility....? but sure!

justinstimatze:
Closing as a duplicate of https://github.com/AIObjectives/tttc-light-js/pull/185 as discussed.

PR #142, opened 2025-04-01
Pipeline steps refactor

Brandon-Perry:
The goal here is to simplify the pipeline steps in express-server/src/pipeline/ , and to use a Result types and custom error classes with them
express-server/src/pipeline/
Ran a report - pending running more 


express-server/src/pipeline/claimsStep.ts
```diff
@@ -1,79 +1,165 @@
...
-      if (statusCode < 200 || statusCode >= 300) {
-        const errorText = await body.json();
-        throw new Error(`Server responded with ${statusCode}: ${errorText}`);
-      }
+      })
+      /**
+       * After we've received the response, clear out the abort signal timer
+       * Brandon: I'm pretty sure this abort signal doesn't do anything since it's not attached to the client. TODO
+       */
+      .then((res) => {
```
staceysv, line 58: this cascade for classifying error types is great — we should wrap all the pipeline steps like this—why is this only in this step?

express-server/src/pipeline/claimsStep.ts
```diff
@@ -1,79 +1,165 @@
...
+    await client.close();
+  }
+}
 
-      // Read the response body
-      const jsonData = await body.json();
+/**
+ * Tests the resposne for error codes we don't like and returns a Result
+ */
+const validateResponse = async (
```
staceysv, line 122: this seems like a more general function — perhaps we could move it into the error-handling logic?

express-server/src/pipeline/claimsStep.ts
```diff
@@ -1,79 +1,165 @@
...
+    return {
+      tag: "success",
+      value: data,
+    };
+  }
+};
 
-      // Validate the response
-      const { data, usage, cost } = apiPyserver.claimsReply.parse(jsonData);
+class InvalidResponseDataError extends CustomError<"InvalidResponseDataError"> {
```
staceysv, line 149: these custom error types are helpful and it seems like we could reuse them across pipeline steps — could we move these three to wherever the error-handling logic lives and import from there to all pipeline steps?

Brandon-Perry: So claimsStep is a special case since we're using this undici client here to prevent timeouts. In handlePipelineStep, you can see we actually do the same thing with `res.ok` on line 20. With claimsStep, we have to manually test the response codes.

express-server/src/pipeline/claimsStep.ts
Brandon-Perry: So my comment below.

express-server/src/pipeline/claimsStep.ts
Brandon-Perry: See my comment below.

express-server/src/pipeline/errors.ts
```diff
@@ -0,0 +1,19 @@
+import { CustomError } from "../error";
+
+export class InvalidResponseDataError extends CustomError<"InvalidResponseDataError"> {
```
staceysv, line 3: do we need multiple copies of these?

staceysv (changes requested):
looks good, left a few questions! also we'll need to rebase this to match the earlier storage PR

staceysv:
re: claimsStep versus others — thank you, I understand now why we have the extra logic for the special/extra-complicated claimsStep relative to the other steps. I'm still not sure about redefining the custom errors in claimsStep.ts.

express-server/src/pipeline/claimsStep.ts
Brandon-Perry: Good catch, that's been fixed.

express-server/src/workers.ts
```diff
@@ -147,15 +155,29 @@ const setupPipelineWorker = (connection: Redis) => {
...
+      // const {
+      //   claims_tree,
+      //   usage: claimsTokens,
+      //   cost: claimsCost,
+      // } = await claimsPipelineStep(env, {
+      //   tree: { taxonomy },
+      //   comments,
+      //   llm: claimsLLMConfig,
+      // });
+      const claimsStep = await claimsPipelineStep(env, {
```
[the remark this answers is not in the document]
Brandon-Perry: The next PR changes this entire file, so lets make sure that happens in the next PR

justinstimatze (approved):
This feels like an ok intermediate step but there's some feedback here that I'd like to see addressed in #146 after it's updated with this merge.

staceysv (approved):

PR #143, opened 2025-04-01
dry_run to save $$/time: skip calls to LLMs, return fixed json

staceysv:
When testing in local mode:
1. Open pyserver/config.py
2. Set DRY_RUN = True
3. Continue testing as you would normally

- mock OpenAI calls on all pyserver routes (topic_tree, claims, sort_claims_tree, cruxes) so we can test the rest of T3C without paying for LLM calls
- finally lint our Python code reasonably with Ruff

- pyserver only to short-circuit each route & return a fixed mock object instead of calling an LLM, if dry_run is passed or set at the config level 

- works locally!

Ideal next steps for when we have time:
- configure DRY_RUN as the default when testing locally
- configure a clean way to turn "calling external LLM" on when working on relevant code paths locally

### DIFF
--- a/pyserver/config.py
+++ b/pyserver/config.py
@@ -112,33 +112,32 @@ return a JSON object of the form
 WANDB_PROJECT_NAME = ""
 WANDB_GROUP_LOG_NAME = ""
 
-DRY_RUN = True
+DRY_RUN = False
 
 MOCK_RESPONSE = {
     "topic_tree": {
-        "data":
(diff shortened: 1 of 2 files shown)


Brandon-Perry (approved):
lgtm

Additionally, a feature where we can add delays to returning the mock data would be good to have. Could be a future PR. Would be good in cases where we want to test the pending state on the frontend

PR #144, opened 2025-04-03
Move all key transport to headers

portiabytes:
This PR standardizes API key handling across T3C by moving all API key authentication to HTTP headers. This improves security by following best practices for sensitive credential handling and ensures consistent authentication across all components.

Schema Changes (common/schema.ts):
- Removed apiKey from llmUserConfig schema
- Removed apiKey from oldOptions schema

Express Server Changes (express-server/src/routes/create.ts):
- Removed API key from request body
- Removed API key password check
- Removed API key from config object

Next.js Client Changes (next-client/src/features/submission/actions/SubmitAction.ts):
- Removed API key from llmUserConfig parsing
- Added API key to request headers using OPENAI_API_KEY_HEADER
- Added necessary import for header constant

Runs locally, have not yet tested end to end to make sure everything is transported correctly because I think we're not using our open API key right now?

### DIFF
--- a/next-client/src/features/submission/actions/SubmitAction.ts
+++ b/next-client/src/features/submission/actions/SubmitAction.ts
@@ -9,6 +9,7 @@ import Papa from "papaparse";
 import { GenerateApiResponse, GenerateApiRequest } from "tttc-common/api";
 import { z } from "zod";
 import { validatedServerEnv } from "@/server-env";
+import * as apiPyserver from "tttc-common/apiPyserver";
 
 const parseCSV = async (file: File): Promise<SourceRow[]> => {
   const buffer = await file.arrayBuffer();
@@ -50,7 +51,6 @@ export default async function submitAction(
   }
   `;
   const config: LLMUserConfig = llmUserConfig.parse({
-    apiKey: formData.get("apiKey"),
     title: formData.get("title"),
     // question: formData.get("question"),
     description: formData.get("description"),
@@ -75,6 +75,7 @@ export default async function submitAction(
     body: JSON.stringify(body),
     headers: {
       "Content-Type": "application/json",
+      [apiPyserver.OPENAI_API_KEY_HEADER]: formData.get("apiKey") as string,
     },
   });
   if (!response.ok) {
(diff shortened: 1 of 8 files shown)


next-client/src/features/submission/actions/SubmitAction.ts
```diff
@@ -75,6 +75,7 @@ export default async function submitAction(
     body: JSON.stringify(body),
     headers: {
       "Content-Type": "application/json",
+      [apiPyserver.OPENAI_API_KEY_HEADER]: formData.get("apiKey") as string,
```
staceysv, line 78: how does this get passed to downstream requests?

staceysv:
Thanks @portiabytes ! Looks good overall!

I'm confused about passing around `env.OPENAI_API_KEY` in the pipeline steps:

In my mind, `env` should be consistent for environment variables/one _deployment_ of T3C like staging versus production, and we want this OpenAI key to potentially be different for every user. I see where we're setting the key in client/src/features/submission/actions/SubmitAction.ts:78, but we're just putting this in the header—how does this propagate to later calls to the four individual steps? I would expect us to save it on config, e.g. as config.apiKey, and then pass that to the pipeline steps? Or, at any rate, not as a variable on env? 

Related note/devops TODO here: we probably DO want to hardcode a single key for the entire staging environment — or have a way for devs to set this in their local envs (for example I have export OPENAI_API_KEY = ... in my pyserver .env.

Minor notes to ponder: 
- eventually we'll want to pass around multiple apikeys (e.g. for Anthropic/Claude in addition to OAI). Maybe we do want to keep a prefix string, like "oaiApiKey", where possible for clarity/for now?
- I understand that `export const OPENAI_API_KEY_HEADER = "X-OpenAI-API-Key";` is probably cleaner/better style, and I would find
```
headers: {
        "Content-Type": "application/json",
        "X-OpenAI-API-Key": oaiApiKey,
```
more immediately readable than:
```
headers: {
        "Content-Type": "application/json",
         [apiPyserver.OPENAI_API_KEY_HEADER]: apiKey,
```
this may be easier to implement after Brandon's refactor lands and makes the four pipeline steps run through one wrapper function so we won't be setting this four separate times, only ones :)

staceysv (changes requested):
looks good to put on staging — let's clarify the env var piece and then go for it!

PR #145, opened 2025-04-03
don't fail on empty comments / speakers in input CSVs

staceysv:
Currently we fail a whole report if any comment is empty. There's also a regression where we fail in workers.ts if a speaker name is not provided / interview column is empty. This draft PR fixes these issues & unblocks immediate generation of reports for CSVs with missing comments or speakers, but there may be more graceful solutions or other edge cases we don't handle yet.
Also added two of my standard test CSVs to examples/sample_csv_data.

- remove requirement for comments to be at least 1 character in schema.ts since users often upload CSVs with some blank rows (empty "comment" fields)
- check for empty comments in pyserver and only call an LLM on one character or more—NOTE, we may want to revisit this and require at least 3-4 characters or even at least a sentence for the extracted claims to be meaningful and not hallucinated
- temporarily remove check on empty speaker in workers.ts — if we do this, we need to do it LATER/after anonymous names are added

Locally

### DIFF
--- a/common/apiPyserver.ts
+++ b/common/apiPyserver.ts
@@ -47,7 +47,7 @@ export type Usage = z.infer<typeof usage>;
  */
 const comment = z.object({
   id: z.string(),
-  text: z.string().min(1, "Empty comment"),
+  text: z.string(),
   speaker: z.string(),
 });
 
--- a/examples/sample_csv_files/missing_fields_pets.csv
+++ b/examples/sample_csv_files/missing_fields_pets.csv
@@ -1,6 +1,8 @@
 id,interview,comment
 1,Dany,
 2,,I love cats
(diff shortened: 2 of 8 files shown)


staceysv:
this is ready to go! 
additional notes:
- assume raw comment contains at least 10 characters—or if not, check for at least 3 words—for us to meaningfully extract claims
- minor: more graceful handling of crux errors (don't assume we can look up a key in case it's empty/we've failed to return it)

TODO/eventually:
- speaker anonymization/tracking may need more thorough testing — I removed a check which perhaps supported some functionality so we don't fail the whole report because of a row
- we will want to modify the "comment is meaningful" check for elicitation/other more conversational modes where a one-word response can be meaningful
- we can do a more sophisticated check on the source quote supporting a claim being present in the raw comment—or at least the length of the full claim being LONGER than the source quote

Brandon-Perry (approved):
lgtm, I can pretty easily reconcile my other two PRs

justinstimatze (approved):
LGTM

PR #146, opened 2025-04-08
Pipeline refactor

Brandon-Perry:
Note: This is based off of #138, might want to preview and merge that first

This is an attempt at simplifying the pipeline on the backend.

Moved most of the code outside of workers.ts and into a new file

The pipeline function uses FP principles and a Result type to make everything more explicit.

I ran the pipeline on the pets dataset (mocked). It would be nice if I could test this on other datasets as well. 


express-server/src/routes/create.ts
```diff
@@ -171,21 +172,41 @@ async function createNewReport(req: Request, res: Response) {
...
-            firebaseJobId: maybeFirebaseJobId,
-          }
-        : null,
+      auth: "public",
+      instructions: {
+        ...userConfig,
+        cruxInstructions: userConfig.cruxInstructions ?? "", // ! Crux instructions should probably be made non-optional?
+      },
+      api_key: "", // ! Change when we transition away from using the AOI key,
+      featureFlags: {
```
justinstimatze, line 194: I don't think I want to call these feature flags, there's some semantic fuzziness/collision. These feel like pipeline job options/parameters.

express-server/src/jobs/pipeline.ts
```diff
@@ -0,0 +1,536 @@
...
+      const { usage, cost } = step.value;
+      const updatedTracker = sumTokensCost({
+        tracker: accum,
+        stepUsage: usage,
+        stepCost: cost,
+      });
+      logTokensInTracker(updatedTracker, step.value.stepName);
+      return updatedTracker;
+    } else {
+      console.log(`TODO`);
```
justinstimatze, line 526: What's the ambiguity if the step fails?

express-server/src/types/result.ts
```diff
@@ -50,6 +54,57 @@ export function mapResult<T, T2, F>(
...
+ *
+ * sequenceResult -> Result<T,F>[] -> Result<T[], F>
+ *
+ * Example:
+ *
+ * sequenceResult([Success(1), Success(2), Success(3)]) -> Success([1,2,3])
+ *
+ * sequenceResult([Success(1), Failure(GoofedError), Success(3)]) -> Failure(GoofedError)
+ *
+ * Devnote: The type stuff here is pretty gnarly. I had Claude generate it because I couldn't figure it out. After a few tries I got this, and it works.
```
justinstimatze, line 80: I'd feel better with some test coverage on this.

express-server/src/jobs/pipeline.ts
```diff
@@ -0,0 +1,536 @@
...
+  };
+  const doCruxStep = async (args: CruxProps) =>
+    await Pyserver.cruxesPipelineStep(env, { ...args, llm: cruxesLLMConfig });
+
+  type Addons = Partial<{
+    crux: CruxProps;
+  }>;
+
+  const doAddons = async (addons: Addons) => {
+    // keep it simple for now, since we really only have one addon
```
justinstimatze, line 425: I have reservations about grouping all addons together, even though yes there's only one right now.

express-server/src/jobs/pipeline.ts
```diff
@@ -0,0 +1,536 @@
...
+  const newTaxonomyResult = mapResult(
+    outputData,
+    ([topicData, _, sortData]) => {
+      const newTax: schema.Taxonomy = topicData.tree.taxonomy.map((t) => ({
+        ...t,
+        topicId: randomUUID(),
+        subtopics: t.subtopics.map((sub) => ({
+          ...sub,
+          subtopicId: randomUUID(),
+          // @ts-ignore // TODO FIX THIS
```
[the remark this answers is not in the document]
Brandon-Perry: It's one of those things where it works, but it's very brittle and not type safe. If there's a change elsewhere in the logic, it could break without warning.

express-server/src/jobs/pipeline.ts
Brandon-Perry: What are your worries? I wasn't 100% sure about having them split off like this. I think if add-ons are always derived from the pipeline steps (and not vice versa), then it should be fine. It would also give us a way to batch promises in the future. But I'm open to changing this.

express-server/src/jobs/pipeline.ts
Brandon-Perry: Oh yeah that console log can be removed.

express-server/src/jobs/pipeline.ts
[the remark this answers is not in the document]
justinstimatze: Made a ticket to get back to it, won't block here.

express-server/src/jobs/pipeline.ts
justinstimatze: I think I misunderstood "addons", but I still have concerns. It seems like in the old code the crux step just returned more data than we usually get from steps and it made some sense to tack it on somewhere else called addons. (Correct me if I'm misinterpreting.) I feel like I'd prefer to keep each step output(s) unambiguously coupled to the step execution, effectively bundling crux addons with the main crux data somehow.

Brandon-Perry: So are you proposing that there's some "step x" that returns either: StepXPipelineOutput | (StepXPipelineOutput & Crux) ? If so, that seems like that violates the single responsibility principle?

express-server/src/types/result.ts
Brandon-Perry: Tests added

justinstimatze: Very nice, thank you!

express-server/src/routes/create.ts
Brandon-Perry: Changed it to options.

justinstimatze (approved):
Good work.

PR #147, opened 2025-04-22
Open Graph for home page. 

Brandon-Perry:

Adds an open graph preview to our next client's metadata so that when its shared on a social media site, it has the correct preview.

root layout of nextjs client

I used a GC plugin to see the previews. Here they are:

image

image

image

**One potential issue: **
The image looks like very top part of the preview image is being clipped. @colleenm How does this look to you? If you want it changed, I think we'll need to change the image itself to be a different aspect ratio. 

Also, I think we should add an open graph preview to each report that's generated. However, I need to change it so that our report data is being fetched server side first. I'm going to throw those tickets into Trello. @justinstimatze 

### DIFF
--- a/next-client/src/app/layout.tsx
+++ b/next-client/src/app/layout.tsx
@@ -3,11 +3,47 @@ import Navbar from "@/components/navbar/Navbar";
 import { Toaster } from "@/components/elements";
 import "./global.css";
 import { getAuthenticatedAppForUser } from "@/lib/firebase/serverApp";
+import { Metadata } from "next";
 
 export const dynamic = "force-dynamic";
 
-export const metadata = {
+export const metadata: Metadata = {
   title: "Talk to the City",
+  openGraph: {
+    title: "Talk to the City",
+    description:
+      "Talk to the City helps large groups of people coordinate by understanding each other better, faster, and in more depth.",
+    siteName: "Talk to the City",
+    images: [
+      {
+        url: "/images/t3c-product-desktop-mobile.png",
+        width: 1200,
+        height: 630,
+        alt: "Talk to the city",
+      },
+    ],
+    locale: "en_US",
+    type: "website",
+  },
+  twitter: {
+    card: "summary_large_image",
+    title: "Talk to the City",
+    description:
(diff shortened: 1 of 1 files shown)


next-client/src/app/layout.tsx
```diff
@@ -3,11 +3,47 @@ import Navbar from "@/components/navbar/Navbar";
...
+  },
+  twitter: {
+    card: "summary_large_image",
+    title: "Talk to the City",
+    description:
+      "Talk to the City helps large groups of people coordinate by understanding each other better, faster, and in more depth.",
+    creator: "@AIObjectives",
+    images: [
+      new URL(
+        process.env.NODE_ENV === "development"
```
justinstimatze, line 36: Will we have to revisit this for staging environments? I assume neither option will be quite right there. Not blocking yet, but noting for potential follow up.

next-client/src/app/layout.tsx
```diff
@@ -27,17 +63,7 @@ export default async function RootLayout({
...
-          // toastOptions={{
-          //   className: "border-none p-4 flex",
-          //   classNames: {
-          //     toast: "bg-accent",
-          //     title: "text-accent-foreground",
-          //     icon: "text-accent-foreground",
-          //   },
-          // }}
-        />
+        <Toaster position="bottom-right" />
```
justinstimatze, line 66: Unrelated? Fine but mildly wish it were in a separate PR.

justinstimatze (approved):
Minor non-blocking opinions, LGTM in general.

next-client/src/app/layout.tsx
Brandon-Perry: Oh this is just removing the commented out code above, idk why it's saying I added that Toaster component.

next-client/src/app/layout.tsx
Brandon-Perry: Hmm so I looked into this, and apparently its convention in the React ecosystem to only use 'development', 'testing', and 'production'. We'll need another way to configure for staging environments. Which I think makes some amount of sense, you want your staging to be a production build right?

colleenm:
@Brandon-Perry cropped image looks fine to me, thanks for checking

PR #149, opened 2025-04-22
Split useReportState into multiple files

Brandon-Perry:
….ts to be more concise

Made the useReportState hook more readable by splitting it into multiple files

Above. 

Changed the tests to be a little simpler to setup

Changed the way actions were defined so we could infer their types instead of writing out every type.

Fixed a bug I found where the subtopic pagination was not being reset correctly

I have a pretty comprehensive set of tests in src/hooks/useReportState/__tests__ that all passed.

Most of the changes are just moving things between files, so I don't expect there to be any new bugs introduced.


next-client/src/components/report/hooks/useReportState/__tests__/testStateSetup.ts
justinstimatze, line 1: I have some mild opinions about splitting these tests into files given the required boilerplate. Is there a distinct advantage to that here?

justinstimatze (commented):
I have minimal opinions on the frontend code here, but I like splitting apart that big file. One open question in which I might learn something, then I'll approve.

next-client/src/components/report/hooks/useReportState/__tests__/testStateSetup.ts
Brandon-Perry: I find big jest / vitest files extremely hard to read when they get too large. I'd much rather have boilerplate than have to find my way around 1000+ lines of nested testing code.

justinstimatze (approved):
LGTM

PR #156, opened 2025-04-29
Restructure tttc-common for dependabot expectations

justinstimatze:

Address this error by restructuring our shared tttc-common module:
Dependabot couldn't fetch all your path-based dependencies. The affected dependencies were tttc-common.

tttc-common, moving files and changing exports as discussed with Brandon.

Tested functionality locally
Tested dependabot locally and this does resolve that error. However there is another new error I can't resolve, but that I think is related to local testing only.


express-server/package.json
Brandon-Perry, line 1: Are you sure this is right? From what I remember, common needs to export the transpiled JS instead of raw TS.

Brandon-Perry (changes requested):

express-server/package.json
Brandon-Perry: Try blasting the node_modules in client and express-server, npm instaling, and running the project. That should answer this.

justinstimatze: I think we're both right? common is exporting the dist folders, so express-server doesn't need to specify that when importing. I tried removing node_modules in client and server, removing dist, rebuilding common, reinstalling in client and server, ran an report successfully.

Brandon-Perry: Oh I see, I missed the /dist/ part in the common package.

PR #169, opened 2025-05-08
T3C-110: CICD main Latest Configs

explomind1:
1. What is the goal of this PR?

- To finalize and activate CI/CD for the production deployment of all T3C components. This enables automatic builds and deployments to Cloud Run whenever a PR is merged into main.

2. What specific parts of T3C are you changing and how?

- Updated .github/workflows/deploy-*.yml files to point to the actual Cloud Run services instead of the test ones. Updated cloudbuild.yaml in each of the three components (next-client, express-server, pyserver) to reflect production build configurations.

- Cleaned up environment variable usage and aligned GitHub Secrets with live services. Removed testing-specific deployment settings (e.g., emre* services) and replaced them with correct production service names.

3. How did you test these changes?

- All CI/CD logic was tested end-to-end in the cicdtrial branch and confirmed to work via GitHub Actions. Verified that changes to any component triggered the correct builds and deployments.

Please use Squash and Merge to keep history clean (I've iterated a lot in this branch).


.dockerignore
```diff
@@ -2,7 +2,18 @@ fixtures/
...
-.env*
+#.env
+#.env*
 **/*.stories.*
-**/.storybook
\ No newline at end of file
+**/.storybook
+
+  
+!next-client/package.json
```
justinstimatze, line 11: I feel like I need the package files in the build context?

.dockerignore
```diff
@@ -2,7 +2,18 @@ fixtures/
...
+!next-client/package-lock.json
+!next-client/Dockerfile
+!next-client/cloudbuild.yaml
+
+
+
+**/firebase-admin.json
```
justinstimatze, line 18: This is useful but surprising and I think means our internal build docs are outdated, since they imply this file is required. Can you double check?

.github/workflows/deploy-pyserver.yml
```diff
@@ -0,0 +1,49 @@
...
+          echo "${{ secrets.ENV_PYSERVER }}" > pyserver/.env
+
+###
+      - name: Build and Push Docker image
+        run: |
+          cd pyserver
+          gcloud builds submit . \
+            --project=tttc-light-js \
+            --config=cloudbuild.yaml \
+            --timeout=1200s
```
justinstimatze, line 39: Timeout seems a little high, a few minutes seems fine. I'd hesitate to block the workflow this long if something does fully timeout.

.gitignore
```diff
@@ -2,7 +2,7 @@ node_modules/
 **/.cache/*
 .DS_Store
 google-credentials.json
-.env*
+#.env*
```
justinstimatze, line 5: I think we do normally want to ignore these?

express-server/.env
```diff
@@ -0,0 +1,17 @@
...
+export OPENAI_API_KEY_PASSWORD = null
+
+export GOOGLE_CREDENTIALS_ENCODED=xxx
+
+export FIREBASE_DATABASE_URL=xxx
+
+export CLIENT_BASE_URL=xxx
+export PYSERVER_URL=xxx
+export GOOGLE_APPLICATION_CREDENTIALS=xxx
+export NODE_ENV=prod
```
justinstimatze, line 17: Should be `production` now.

express-server/Dockerfile
```diff
@@ -14,7 +12,7 @@ COPY express-server/package*.json ./express-server/
 WORKDIR /usr/src/app/common
 RUN npm install --only=production
 RUN npm install typescript @types/node vite rollup @rollup/plugin-node-resolve @rollup/plugin-commonjs rollup-plugin-typescript2 --save-dev
-
+
```
justinstimatze, line 15: Just whitespace changes in this file?

justinstimatze (changes requested):
I really like most of this, thank you! Some minor questions and nitpicks, plus it needs a rebase because of some Dockerfile conflicts, but we're almost there.

.dockerignore
explomind1: Yes I just made sure that they are included with "!" to prevent .dockerignore from ignoring it by default sometimes.

.dockerignore
explomind1: I reviewed the Dockerfiles and the Firebase.ts module, and it looks like firebase-admin.json isn’t used during the build — the app uses applicationDefault() which depends on runtime credentials (like the GOOGLE_APPLICATION_CREDENTIALS env variable). So from what I can see, .dockerignore-ing it is safe. That said, if there's anything I might be overlooking, I'm happy to revisit it.

.gitignore
explomind1: I kept on hitting errors when ignoring .env, so I kept placeholder .env files to unblock CICD setup. It was a quick fix to stay focused, but happy to revisit if we want a cleaner solution.

.github/workflows/deploy-pyserver.yml
explomind1: I agree just changed it!

express-server/Dockerfile
explomind1: No I am not sure why its only showing like that- because if I remember correctly my changes were the following: 
The block does include important changes I made to fix Cloud Build errors — like installing dev dependencies(I added the line that installs typescript, rollup, and related dev dependencies) for common and express-server that weren't resolving during CI/CD builds.
The whitespace came from debugging and iterating during that fix. I'm also not entirely confident about the differences between the express-server Dockerfile in main and my branch(cicdtrial) — or if they might lead to bigger issues down the line. My priority was getting the build to work (it was a bit of a rabbit hole, so I tried a few variations). This version does build successfully, but let me know if there’s anything that looks unsafe or needs adjustment.

.dockerignore
justinstimatze: Apologies, I misunderstood.

justinstimatze (approved):
LGTM, good work. Excited to have this in place.

PR #177, opened 2025-05-15
T3C-245: Fix missing Firebase credentials by deprecating GOOGLE_APPLICATION_CREDENTIALS

justinstimatze:

Address a runtime bug after CICD builds where the Firebase credentials file referred to by GOOGLE_APPLICATION_CREDENTIALS is not present in the image. We now check a new var FIREBASE_CREDENTIALS_ENCODED for the same content.

Environment variable handling.

Manually, locally and in staging.


Brandon-Perry (dismissed):

Brandon-Perry:
@justinstimatze 

According to the Firebase SDK docs, it looks like it initializing everything with this way in non-Google environments. 
https://firebase.google.com/docs/admin/setup#initialize_the_sdk_in_non-google_environments

Above that section, it goes into using a env var for adding the credentials. So maybe we need to do this one way for dev, and another for staging/prod?

justinstimatze:
It's useful to have these as env vars managed by a secrets manager (ideally it would be mounted paths, but we'll get to that). I tried using the same credentials for both buckets and firestore, but that's not at all how the existing service accounts are set up and maybe that's fine. So, rather than mix the semantics on the existing GOOGLE_APPLICATION_CREDENTIALS (a file path), I've added a new env var FIREBASE_CREDENTIALS_ENCODED that clarifies what those credentials are used for and what their format is. This should work similarly locally and in production.

PR #178, opened 2025-05-16
adding basic test cases in example CSV files

staceysv:
- add some basic test cases for making sure comments are parsed correctly

- Consolidating and renaming some test CSV files in examples/

- Locally — note that the hard edge cases will cause the pipeline to fail right now because they yield empty subtopics. 

### DIFF
--- a/examples/sample_csv_files/test_data/test_hard_cases_22.csv
+++ b/examples/sample_csv_files/test_data/test_hard_cases_22.csv
@@ -1,0 +1,23 @@
+id,interview,comment
+1,Dany,
+2,,I love cats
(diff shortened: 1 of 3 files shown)


justinstimatze (approved):
Seems reasonable, but I wish there were a way to comment in CSVs to indicate what the test cases actually are. Some I can guess the intention, but I assume I'm missing some too. Perhaps when these are primarily called from API tests, we can annotate in the calling functions.

staceysv:
thanks, that's a great point — I have a partial "taxonomy of test cases" written out & will complete it in the manual QA notes. Agree having this very explicit/enumerated somewhere/commented directly in the test cases would be ideal.

PR #179, opened 2025-05-17
T3C-249: Change Firebase config from buildtime to runtime

justinstimatze:

Change next-client's Firebase config from buildtime to runtime so we don't need to specify those environment variables during the build using values for production.

next-client firebase config

Local report browsing and generation


justinstimatze:
This was easier to test building off the T3C-245 branch but I'll fix the history after that's merged.

Brandon-Perry (approved):

PR #180, opened 2025-05-20
T3C-249: Refine firebase usage across client and server

justinstimatze:

Continuing the work in https://github.com/AIObjectives/tttc-light-js/pull/179 because I missed some things in my testing. We want to be able to build next-client without requiring the Firebase credentials to be specified at that time, but then able to specify them at runtime to the container. 

This led to a rabbit hole of how Firebase is handled differently between the client and admin SDKs, so that the right credentials are dynamically used at the right time in the right places.

next-client firebase handling, both client and server.

Local report generation in dev mode with an environment, successful container build without an environment.

### DIFF
--- a/next-client/src/app/my-reports/page.tsx
+++ b/next-client/src/app/my-reports/page.tsx
@@ -1,11 +1,13 @@
 "use client";
 import { useUser } from "@/lib/hooks/getUser";
-import { getUsersReports } from "../../lib/firebase/firestore";
-import { db } from "@/lib/firebase/clientApp";
+import { getUsersReports } from "../../lib/firebase/firestoreClient";
+import { getFirebaseDb } from "@/lib/firebase/clientApp";
 import { useAsyncState } from "@/lib/hooks/useAsyncState";
 import { useEffect, useState } from "react";
 import MyReports from "@/components/myReports/MyReports";
 import { Spinner } from "@/components/elements";
+
+const db = getFirebaseDb();
 
 function Center({ children }: React.PropsWithChildren) {
   return (
--- a/next-client/src/components/feedback/Feedback.tsx
+++ b/next-client/src/components/feedback/Feedback.tsx
@@ -15,6 +15,7 @@ import { feedbackResponse } from "@/lib/types/clientRoutes";
 import { toast } from "sonner";
 import Icons from "@/assets/icons";
 import { cn } from "@/lib/utils/shadcn";
+import { fetchToken } from "@/lib/firebase/getIdToken";
 
 export default function Feedback({ className }: { className?: string }) {
   return (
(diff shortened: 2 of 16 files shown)


next-client/src/lib/firebase/clientApp.ts
Brandon-Perry, line 1: Could you add some comments to the Firebase code in general? Just explaining what it does and why. When I originally implemented everything I had to do it quickly. So, I don't quite remember how everything is supposed to work.

next-client/src/components/feedback/Feedback.tsx
```diff
@@ -85,16 +101,16 @@ function FeedbackForm() {
       })
       .then(async (res) => {
         const json = await res.json();
+        const response = feedbackResponse.parse(json).response;
         console.log("json", json);
-        return json;
```
Brandon-Perry, line 89: I'm wondering what prompted these changes here?

next-client/src/app/my-reports/page.tsx
```diff
@@ -58,8 +60,9 @@ function MyReportsUI({ userId }: { userId: string }) {
     );
   const reports = result[1];
 
+  // MyReports only needs to be horizontally centered.
   return (
-    <div className="justify-items-center">
```
Brandon-Perry, line 62: Related?

Brandon-Perry (changes requested):
Mostly good, but there's a few parts where I'd suggest changes / confused about something.

next-client/src/app/my-reports/page.tsx
justinstimatze: After I modified layout.tsx, this was the only thing that appeared to break. But if it's actually from something else or you want to fix it separately then I can remove this.

next-client/src/app/my-reports/page.tsx
```diff
@@ -1,15 +1,17 @@
...
 import { useEffect, useState } from "react";
 import MyReports from "@/components/myReports/MyReports";
 import { Spinner } from "@/components/elements";
 
+const db = getFirebaseDb();
+
 function Center({ children }: React.PropsWithChildren) {
   return (
-    <div className="w-full h-full content-center justify-items-center">
+    <div className="flex justify-center items-center w-full h-full">
```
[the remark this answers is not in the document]
justinstimatze: See https://github.com/AIObjectives/tttc-light-js/pull/180/files#r2103458757

next-client/src/components/feedback/Feedback.tsx
```diff
@@ -72,8 +73,23 @@ function FeedbackForm() {
...
+      if (Array.isArray(tokenResult) && tokenResult[0] === "data") {
+        token = tokenResult[1];
+      } else {
+        token = null;
+      }
+    }
+    const headers: Record<string, string> = {
+      "Content-Type": "application/json",
+    };
+    if (token) headers.Authorization = `Bearer ${token}`;
```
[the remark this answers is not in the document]
justinstimatze: No, it's Firebase token handling because that's how Firebase validates user requests from the client. Unrelated to the API headers PR.

next-client/src/components/feedback/Feedback.tsx
justinstimatze: Just complications from adjusting the code above, but mixing 'response' and 'res' is confusing. My impression is that this component probably needed to be revisited later.

next-client/src/components/navbar/components/LoginButton.tsx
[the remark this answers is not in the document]
justinstimatze: Good catch, it does not. I'll update the logic but keeping the changes small for this PR has some tradeoffs. I've made T3C-285 to handle a more general approach.

next-client/src/components/feedback/Feedback.tsx
justinstimatze: I've tried to clarify this a little with another pass.

Brandon-Perry (approved):
lgtm

PR #182, opened 2025-05-22
T3C-101: Move OpenAI API keys to headers

justinstimatze:

This is a manual recreation of #144 which is a subset of #99, due to a messy rebase across a big refactor. Functionality is very similar, but I am ignoring the changes related to user-provided keys for now. That will be a future change once fully scoped.

OpenAI API key handling betwen express-server, common, and pyserver, from params to headers.

Local report creation, with and without valid keys in the relevant environments.


pyserver/main.py
```diff
@@ -577,10 +579,17 @@ def all_comments_to_claims(
     return {"data": node_counts, "usage": net_usage, "cost": s2_total_cost}
 
 
-def dedup_claims(claims: list, llm: LLMConfig) -> dict:
-    """Given a list of claims for a given subtopic, identify which ones are near-duplicates
+def dedup_claims(claims: list, llm: LLMConfig, api_key: str) -> dict:
```
Brandon-Perry, line 582: How is this related to the PR?

pyserver/main.py
```diff
@@ -1104,11 +1113,12 @@ def controversy_matrix(cont_mat: list) -> list:
 
 
 def cruxes_for_topic(
-    llm: dict, topic: str, topic_desc: str, claims: list, speaker_map: dict,
+    llm: dict, topic: str, topic_desc: str, claims: list, speaker_map: dict, api_key: str
 ) -> dict:
     """For each fully-described subtopic, provide all the relevant claims with an anonymized
     numeric speaker id, and ask the LLM for a crux claim that best splits the speakers' opinions
-    on this topic (ideally into two groups of equal size for agreement vs disagreement with the crux claim)
+    on this topic (ideally into two groups of equal size for agreement vs disagreement with the crux claim).
```
Brandon-Perry, line 1120: Looks like you have a crux PR mixed in with this?

Brandon-Perry (approved):
Looks like there's some changes related to cruxes mixed in here. Not a big deal imo if you don't want to separate it out, your call. Other than that looks good.

pyserver/main.py
justinstimatze: It needs the API key to call OpenAI, correct me if I'm wrong.

pyserver/main.py
justinstimatze: I'm not sure what you mean, can you elaborate?

pyserver/main.py
Brandon-Perry: Oh I see my bad!

pyserver/main.py
Brandon-Perry: I'm just wondering why there are changes to the prompt when this PR is about api headers

justinstimatze: Oh I see what you mean, but I think this is just updating the docstring? The prompt comes in through the llm argument. A new parameter was added (api_key), although I admit this format is non-standard.

PR #183, opened 2025-05-23
Playwright tests

Brandon-Perry:

sets up Playwright and our initial suit of testing. We can always add more tests from here, but I think this is a good start. It's also only setup for local testing. It probably won't work in CI.

next-client/playwright is setup. Also I added some data-testid tags to some of our components. Those are stripped out in prod, so it shouldn't affect anything else.  

The tests pass. Now there's a little more to this. When you run a test suit, you might have 2 or 3 tests fail because of timeout errors. I increased the timeout period, but that only sometimes seems to help. If you're unsure, use `test.only` on the ones that failed your previous run and see if they work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ __pycache__
 /utils/**/input/*
 !**/*/.keep
 */tsconfig.tsbuildinfo
+playwright-report
+test-results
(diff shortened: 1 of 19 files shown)


justinstimatze (approved):
Nice start, thank you! I'd like some follow-ups on the flaky tests and adding to CI, but those can come later.

PR #185, opened 2025-05-27
T3C-71 crux extraction should be optional

Brandon-Perry:

Adds a switch to our report creation form that toggles cruxes

Added switch to form, enabled on backend

Ran pets report. However, the returned data structures were empty. @staceysv is there a dataset that you use specifically to test cruxes?


staceysv:
so excited to review this, thank you!!! @Brandon-Perry try this file that's in the repo, it should have one crux? `main/examples/sample_csv_files/test_data/test_scifi_10.csv`

express-server/src/jobs/pipeline.ts
```diff
@@ -73,7 +73,7 @@ export async function pipelineJob(job: Job<PipelineJob>) {
   const { data, config, reportDetails } = jobdata;
   const { auth, env } = config;
   const { title, description, question, filename } = reportDetails;
-
+  // console.log('config', config)
```
justinstimatze, line 76: Seems like a leftover.

staceysv:
@Brandon-Perry @justinstimatze I can't get this to work on my setup - tried adding a lot of console.log() and a few different ways of checking the value of cruxes / cruxesEnabled. As far as I can tell, when I toggle the switch
- A: the UI state is correct in next-client / SubmitAction.ts  (`config.cruxesEnabled` and `body.userConfig.cruxesEnabled` get set to `true`)
- B: the resulting state of the crux step in express-server, for both `create.ts` and `pipeline.ts`, is `undefined`:  `config.options.cruxes` and updatedConfig.cruxesEnabled are always `undefined`

Here's the terminal output where you can compare, first request no cruxes, second with cruxes.

What am I missing? Is my setup different? It seems like we're just not passing the correct state from next-client to express-server. Note that if I hardcode the cruxes to true, we DO call the pyserver cruxes route and everything is correct :)

Screenshot 2025-05-27 at 18 01 04

staceysv (changes requested):
Left a few comments and added a reliable test file — let's make sure we can:

P0
1. Upload that test file, leave crux extraction unchecked, should NOT run the pyserver/crux route / should return empty addOns object in JSON download
2. Upload that test file, enable crux extraction, should run the pyserver/crux route, should return crux details in addOns object in JSON download

P1:
3. move the whole section about "Enable research features" INSIDE of the "Show advanced settings" menu — do not show it by default.

staceysv:
confirming that this works correctly for me after resetting my dev environment!

justinstimatze (approved):
Minor non-blocking comment, overall LGTM.

staceysv (approved):
thank you, works great for me! I pushed a small change to the user instructions to dedupe/improve for clarity

PR #197, opened 2025-06-20
GitHub actions local

Brandon-Perry:
Note: this is currently based off of PR #196 

There's a tool you can use to locally test github actions called [act](https://nektosact.com/introduction.html) I added a line to each of the deploy jobs that prevents it from running when we're using act.

I think in the future we could instead use a dry run instead when we're testing locally.

Changes github actions, adds an .actrc file to root

I ran act and can see that its running correctly.


justinstimatze (approved):
Great, glad to have this ability.

PR #205, opened 2025-07-14
T3C-372 set next client to use strict mode

Brandon-Perry:

Makes the next client use strict mode

Also removed two old files that weren't used in the code base. They would have required changes due to the strict mode change, but figured I might as well just remove them instead.


justinstimatze (approved):
LGTM

PR #206, opened 2025-07-14
T3C-374 Change async data

Brandon-Perry:

Made it so that the AsyncState type + hook uses the Result Type


justinstimatze (approved):
Much cleaner, thanks!

PR #227, opened 2025-08-07
T3C-431: Fix bug in user account creation

justinstimatze:
Additional logging around ensuring user documents

Address a bug that was causing some user account documents to not be created on login.

Frontend user document ensure flow.

Tests and local login testing.


next-client/src/components/navbar/components/LoginButton.tsx
```diff
@@ -81,14 +81,21 @@ export default function LoginButton() {
       ) : (
         <DropdownMenu>
           <DropdownMenuTrigger className="flex items-center">
-            <Avatar className="h-10 w-10 self-center">
+            <Avatar
+              className={`h-10 w-10 self-center ${userDocumentError ? "ring-2 ring-yellow-500" : ""}`}
```
Brandon-Perry, line 85: Since we're changing UI behavior here we should make sure to run this past Pawel first.

next-client/src/components/create/CreateReport.tsx
```diff
@@ -29,7 +29,7 @@ import {
  * Hook for fetching the user's OAuth token. Needed before report form is shown.
  */
 function getUserToken(): AsyncState<string | null, Error> {
-  const { user, loading } = useUser();
+  const { user, loading, userDocumentError } = useUser();
```
Brandon-Perry, line 32: userDocumentError isn't being used here

justinstimatze: Good catch, thanks.

next-client/src/lib/hooks/getUser.ts
```diff
@@ -45,12 +49,38 @@ export function useUser() {
...
               const uid = authUser.uid; // Store UID in a local variable to avoid race conditions
               ensurePromise = ensureUserDocumentOnClient(authUser)
-                .then(() => {
-                  ensuredUsersRef.current.add(uid);
+                .then((result) => {
+                  if (!mounted) return;
+
+                  if (result.tag === "success") {
+                    logger.info(
+                      `User document ensured successfully for ${uid}`,
```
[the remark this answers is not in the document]
justinstimatze: Reasonable! Will fix. We're in a bit of a transition on logging conventions and should probably standardize soon.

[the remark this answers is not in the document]
juggler434: Yeah, I wasn't sure what kind of standardization was expected which is why I just left it as a comment instead of a change request. LGTM now.

juggler434 (approved):

PR #229, opened 2025-08-12
T3C-325 Waitlist implementation

Brandon-Perry:

This implements the wait-list check when the user tries to log in. If they are wait-listed,  it will immediately log them out and show a modal alerting them.

Changes ensureDocument to return the user object. Will then do the check in the auth logic. 

I made it so one of my emails is wait-listed, and tried to log in. 


justinstimatze (approved):
This looks solid to me but doesn't change the default value of isWaitlistApproved for new users to invoke this logic. I'm fine with that being a separate, feature enabling change if that's what you intended so not blocking here and just mentioning.

PR #230, opened 2025-08-13
T3C-421 Updates port mapping for production python server

juggler434:
Locally the python server is running at 8000 but in production is running on port 8080.  This will ensure that local and production environments are running on the same ports.
This changes the port that is being listened on for the Python service.  There should be noticeable effect.  
Ran local docker-compose to ensure that pyserv was listening on port 8000

### DIFF
--- a/pyserver/Dockerfile
+++ b/pyserver/Dockerfile
@@ -7,9 +7,9 @@ RUN pip install --no-cache-dir --upgrade -r requirements.txt
 RUN useradd --no-create-home appuser
 COPY --chown=appuser pyserver/ .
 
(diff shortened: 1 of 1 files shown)


juggler434:
> LGTM but I remember last time I worked on this thinking that last port number should reference the $PORT variable. Up to you, non-blocking.

Good point.  I updated the PR.  For our slack conversation about removing the host and port from the dockerfile, apparently when using the uvicorn main:app from the command line it doesn't use main.py startup procedure (according to Claude), so you have to specify.

justinstimatze (approved):
LGTM

PR #250, opened 2025-08-20
Reformat

justinstimatze:

Running npm run format to catch some missed files.

Formatting.

No changes to test.


Brandon-Perry:
Format should have been running on every commit via Husky. I think we should fix that in here along with the other changes.

justinstimatze:
Addressing that in T3C-488, want to keep pure format changes separate from functional changes.

PR #259, opened 2025-08-22
T3C-476 Implements analytics using posthog or local printing

juggler434:
Implement an Analytics Provider for both client and server.
This adds an analytics provider than can be initialized either to print to logs or sent to Posthog
Tracked user sign in events using both providers

### DIFF
--- a/common/analytics/__tests__/helpers.test.ts
+++ b/common/analytics/__tests__/helpers.test.ts
@@ -1,0 +1,531 @@
+/**
+ * Tests for analytics helper functions from index.ts
+ */
+
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import {
+  createAnalyticsConfig,
+  createEventProperties,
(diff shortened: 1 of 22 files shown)


common/analytics/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,533 @@
...
+        // Avatar variations
+        photoURL: "https://example.com/photo.jpg",
+        avatar: "https://example.com/avatar.jpg",
+
+        // Timestamp variations
+        createdAt: new Date("2023-01-01"),
+        lastLoginAt: new Date("2023-06-01"),
+
+        // Plan/subscription variations
+        plan: "premium",
```
justinstimatze, line 410: I don't think we really have these but I also don't think it hurts

justinstimatze (approved):
Minor comment, looks reasonable to me!

PR #265, opened 2025-08-26
T3C-525 made it so tests use ai assemblies data instead of heal michigan

Brandon-Perry:
So I looked a bit more into the tests and why they were failing, and I realized that there are certain aspects of the original ai assemblies data that are needed. For example, one of the failing tests had to do with applying 'open' repeatedly. However, that particular test topic/subtopic didn't have enough children to really test that. Rather than have to redo the indexes for picking good test topics and such, it would be better to add AI assemblies back into the repo to be used for test data.

All tests on the next-client are now passing

Looking at the tests a bit more, I think it would be worth refactoring them to not depend on any particular data at some point in the future. That would hopefully make it easier to make changes, extend tests, etc.


next-client/src/components/report/hooks/useReportState/__tests__/useReportState.increment.test.ts
```diff
@@ -73,7 +67,7 @@ describe("Subtopic Node", () => {
   const newState2 = incrementSubtopic(newState1, getTestSubtopic(state).id);
 
   describe("Precheck", () => {
-    test("Make sure our test subtopic's children length is >= 10", () => {
+    test("Make sure our test subtopic's children legnth is >= 10", () => {
```
justinstimatze, line 70: Typo unfixed

PR #270, opened 2025-08-30
T3C-491: Remove unnecessary .env files from deploy workflows

justinstimatze:

Remove unnecessary .env files and logic from build and deploy scripts. Only next-client requires a buildtime .env, the rest are runtime via CloudRun.

Build and deploy workflows for express-server and pyserver.

Hard to test without actually redeploying staging, so untested.

### DIFF
--- a/.github/workflows/deploy-express-server-production.yml
+++ b/.github/workflows/deploy-express-server-production.yml
@@ -36,10 +36,6 @@ jobs:
 
       ####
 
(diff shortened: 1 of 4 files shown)


explomind1 (approved):
looks fine to me!

juggler434 (approved):
I was just looking at these scripts today and thinking "hey, I don't think this part does anything!"

PR #284, opened 2025-09-10
T3C-169: Add large uploads user permission

justinstimatze:

Adds permissions for some users to be able to allow slightly larger inputs.

Added permissions system, modified client and server upload logic.

Local report generation.

### DIFF
--- a/next-client/src/components/create/hooks/useUserCapabilities.ts
+++ b/next-client/src/components/create/hooks/useUserCapabilities.ts
@@ -1,0 +1,75 @@
+import { useState, useEffect } from "react";
+import { getUserCapabilities } from "@/lib/api/userLimits";
+import { toast } from "sonner";
+import pRetry from "p-retry";
+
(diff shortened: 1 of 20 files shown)


next-client/src/components/create/hooks/useUserCapabilities.ts
```diff
@@ -0,0 +1,75 @@
...
+          const result = await getUserCapabilities();
+          return result;
+        },
+        {
+          retries: 3,
+          minTimeout: 1000,
+          maxTimeout: 8000,
+          onFailedAttempt: (error) => {
+            console.log(
+              `Retrying capabilities fetch (attempt ${error.attemptNumber}/${error.retriesLeft + error.attemptNumber}) - ${error.message}`,
```
juggler434, line 35: I think you should be able to use a logger here to get some formatting for free

juggler434 (changes requested):

PR #300, opened 2025-09-24
T3C-451 Replace BullMQ with Google PubSub

juggler434:
Replace the BullMQ implementation with a google Pubsub solution
Queue reading/writing
Through local emulation, unit tests, and ephemeral environment


express-server/src/workers.ts
```diff
@@ -1,84 +1,92 @@
...
-            error: updateError,
-            originalJobError: e,
-            jobId: job?.id,
-          },
-          "Failed to update Firebase REPORT_REF to failed status",
-        );
-        // TODO: do we want to throw an error here?
-        // throw new Error("Could not update Firestore REPORT_REF to failed status: " + updateError.message)
-      }
+export async function processJobFailure(job: PipelineJob, err: any) {
```
justinstimatze, line 30: Does this need to be an any type?

express-server/src/queue/providers/googlePubSub.ts
```diff
@@ -0,0 +1,94 @@
...
+    );
+    this.subscription.on("message", async (message: Message) => {
+      const jobData = JSON.parse(message.data.toString()) as PipelineJob;
+      try {
+        await processJob(jobData);
+        message.ack();
+      } catch (error) {
+        message.nack();
+        await processJobFailure(jobData, error);
+        throw error;
```
justinstimatze, line 82: Does this also stop listening after a job failure?

justinstimatze (approved):
Seems reasonable, but I think I can't check the ephemeral env because of the new env vars?

juggler434:
> Seems reasonable, but I think I can't check the ephemeral env because of the new env vars?

That's weird, the ephemeral environment is working for me.  The environment variable should be getting set in deploy/cloudrun/express-server.yaml

What are you seeing when you try?

justinstimatze:
> What are you seeing when you try?

On that attempt, it stayed stuck at queued. But, trying again just now, it works fine. Could have been a cold start and I was impatient.

PR #306, opened 2025-10-02
T3C-630 Allow more than one report to be created at a time

juggler434:
Allow multiple reports to be generated at once
Updating the listen command to work with a promise chain instead of await
ephemeral env


justinstimatze (approved):
Seems reasonable to me.

PR #311, opened 2025-10-06
T3C-327: Allow flexible column names to support CSV extractor

justinstimatze:

Address the difference between default outputs of the csv extractor and the default input format. 

CSV parsing, mostly in express-server via common.

Localy report generation using exports from csv extractor.

### DIFF
--- a/common/utils/index.ts
+++ b/common/utils/index.ts
@@ -1,5 +1,77 @@
+import { SourceRow } from "../schema";
+
 export function assertNever(x: never): never {
   throw new Error("Unexpected object: " + JSON.stringify(x));
+}
+
+/**
+ * Formats raw CSV data with flexible column name mapping
+ * Supports multiple CSV formats (WhatsApp, standard T3C, etc.)
+ * @param data - Raw parsed CSV data with any column names
+ * @returns Formatted data conforming to SourceRow schema
+ */
+export function formatData(data: Record<string, unknown>[]): SourceRow[] {
+  // Flexible column name mapping for different CSV formats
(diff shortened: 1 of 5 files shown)


common/utils/index.ts
```diff
@@ -1,7 +1,79 @@
...
+  ];
+
+  if (!data || !data.length) {
+    throw Error("Invalid or empty data file");
+  }
+  const keys = Object.keys(data[0]);
+  const lowerKeys = new Map(keys.map((k) => [k.toLowerCase(), k]));
+
+  const id_column = ID_COLS.map((col) => lowerKeys.get(col.toLowerCase())).find(
+    (x) => x,
```
juggler434, line 43: not crazy about using x here.  By the time I've gotten here in this function chain I'm not sure what x is supposed to be.

common/utils/index.ts
```diff
@@ -1,7 +1,79 @@
...
+  );
+  const comment_column = COMMENT_COLS.map((col) =>
+    lowerKeys.get(col.toLowerCase()),
+  ).find((x) => x);
```
juggler434, line 47: same

common/utils/index.ts
```diff
@@ -1,7 +1,79 @@
...
+  const interview_column = INTERVIEW_COLS.map((col) =>
+    lowerKeys.get(col.toLowerCase()),
+  ).find((x) => x);
```
juggler434, line 50: same, also would be nice if the formatting matched on these 3 calls.

juggler434 (changes requested):
Small nit, other than that lgtm

PR #312, opened 2025-10-07
T3C-640: Clarify CSV filtering involving at-symbol

justinstimatze:
Extend this to plus and minus symbols as well.

The at-symbol filter was too broad, causing input validation problems for valid input.

common

Local report generation with previously failed input.


juggler434 (approved):
LGTM, ran a report with an @ symbol in name and worked as expected

PR #324, opened 2025-10-14
T3C-523 added server-side analytics to nextjs, logs about and home page

Brandon-Perry:
Adds server side analytics to the next client, will log when the user visits the about and home pages.

Client side analytics will be added later.

@juggler434 Can you double check to make sure this is working correctly? I do see some logs and I think it works, but I can't test it with posthog proper.

### DIFF
--- a/next-client/src/lib/analytics/serverSideAnalytics.ts
+++ b/next-client/src/lib/analytics/serverSideAnalytics.ts
@@ -1,0 +1,25 @@
+"use server";
+
+import {
+  createAnalyticsConfig,
+  AnalyticsConfig,
+  initializeAnalytics,
+  getAnalytics,
+} from "tttc-common/analytics";
+
+const config: AnalyticsConfig = createAnalyticsConfig(
+  process.env.NODE_ENV === "development" ? "local" : "posthog",
+  process.env.ANALYTICS_API_KEY,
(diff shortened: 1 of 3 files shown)


next-client/src/lib/analytics/serverSideAnalytics.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const config: AnalyticsConfig = createAnalyticsConfig(
+  process.env.NODE_ENV === "development" ? "local" : "posthog",
+  process.env.ANALYTICS_API_KEY,
+  {
+    host: process.env.ANALYTICS_HOST,
+    enabled: Boolean(process.env.ANALYTICS_ENABLED),
+    flushAt: Number(process.env.ANALYTICS_FLUSH_AT),
+    flushInterval: Number(process.env.ANALYTICS_FLUSH_INTERVAL),
+  },
```
juggler434: These environment variables will probably need to be added in Github Actions/GCP.  I have express-server set to disabled for everything except prod and just turn it on if I need to test something so that testing data isn't polluting our metrics

juggler434 (approved):
After adding in the necessary environment variables on the ephemeral environment I was able to see the metrics on posthog

PR #333, opened 2025-10-20
T3C-660 pipeline worker repo init

Brandon-Perry:
Bare bones setup for the pipeline workers area. I only added a few folders based on the few tickets that were there.


juggler434 (approved):
LGTM

PR #342, opened 2025-10-22
T3C-505: Add legacy report migration script

justinstimatze:

Add a tool for handling legacy or demo reports that may not have reportRef/ids so we can have id based urls for them.

express-server utils

Local dry runs.

### DIFF
--- a/utils/src/scripts/migrate-legacy-report.ts
+++ b/utils/src/scripts/migrate-legacy-report.ts
@@ -1,0 +1,505 @@
+#!/usr/bin/env tsx
+
+/**
+ * Migration script to create Firestore ReportRef entries for legacy reports
+ *
+ * This script migrates reports that exist in Google Cloud Storage but lack
+ * Firestore database entries. Use this for:
+ * - Demo reports displayed on talktothe.city homepage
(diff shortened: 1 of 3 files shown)


explomind1 (approved):
look fine to me.

utils/src/scripts/migrate-legacy-report.ts
```diff
@@ -0,0 +1,458 @@
...
+} as const;
+
+// Schema version for legacy reports
+const LEGACY_REPORT_SCHEMA_VERSION = 1;
+
+const getCollectionName = (env: Env, name: keyof typeof COLLECTIONS) => {
+  return env.NODE_ENV === "production"
+    ? COLLECTIONS[name]
+    : `${COLLECTIONS[name]}_dev`;
+};
```
juggler434: With only one collection available this is probably more than we need here.  We could probably just set the collection name as a constant and append _dev if it's not production.

juggler434 (changes requested):
The migrateReport function is really long, would be nice to have that broken up into a few different functions.  

There's some inconsistency around whether things are returning/throwing and letting the top level call end the process in it's .then().catch block or whether process.exit() is being called in the function itself.

PR #349, opened 2025-10-23
T3C-661 Adds docker file for Node worker

juggler434:
Create a Dockerfile so for the Node worker so an image can be created in GCP and deployed
New node pipeline worker
Created an image locally


justinstimatze (approved):
No notes.

PR #372, opened 2025-11-13
T3C-568: Enable email logins

justinstimatze:

Enable non-Google account logins.
Switch from popup login to redirect flow login.

next-client, most of the login flow.

Testing in ephemeral because of email sending requirement.

### DIFF
--- a/next-client/src/app/auth/action/page.tsx
+++ b/next-client/src/app/auth/action/page.tsx
@@ -1,0 +1,304 @@
+"use client";
+
+import { useEffect, useState, useRef, useCallback, Suspense } from "react";
+import { useSearchParams } from "next/navigation";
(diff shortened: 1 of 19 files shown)


next-client/src/app/auth/action/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+    // For password reset, add query param to auto-open sign-in modal
+    const destination =
+      mode === ACTION_MODES.RESET_PASSWORD
+        ? `/?action=${AUTH_ACTIONS.SIGNIN}`
+        : "/";
+    console.debug("[auth-action] Navigating to home page", { destination });
+    window.location.href = destination;
+  };
+
+  const renderContent = () => {
```
juggler434: This function is kind of long.  The <div>s could probably be pulled into constants and this function would be more readable.

PR #373, opened 2025-11-13
T3C-746: Enable bridging content with Perspective API

justinstimatze:

Adopt the Jigsaw Perspective API for evaluating content for bridging qualities.

Added to most of pipeline.

Locally.

### DIFF
--- a/next-client/src/components/claim/Claim.tsx
+++ b/next-client/src/components/claim/Claim.tsx
@@ -1,14 +1,24 @@
-import React from "react";
+import React, { useContext } from "react";
 import Icons from "@/assets/icons";
 import { Row, Col } from "../layout";
 import * as schema from "tttc-common/schema";
 import { getQuotes } from "tttc-common/morphisms";
 import { InteractiveQuoteCard } from "./HoverQuoteCard";
 import { CopyLinkButton } from "../copyButton/CopyButton";
-import { getThemeColor } from "@/lib/color";
 import { useThemeContextColor } from "@/lib/hooks/useTopicTheme";
-import config from "tailwind.config";
 import { Quote } from "../quote/Quote";
+import { ReportContext } from "../report/Report";
+import {
+  getClaimBridgingScore,
+  BRIDGING_THRESHOLDS,
+} from "@/lib/bridging/utils";
+import {
+  HoverCard,
+  HoverCardContent,
+  HoverCardPortal,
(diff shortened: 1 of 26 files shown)


next-client/src/components/claim/Claim.tsx
```diff
@@ -64,3 +77,49 @@ export function QuoteIcon({ num }: { num: number }) {
...
+  score,
+  forceShow = false,
+}: {
+  score: number;
+  forceShow?: boolean;
+}) {
+  // Don't show badge for neutral scores (between 0 and 0.3) unless forceShow
+  if (!forceShow && score >= 0.0 && score < 0.3) {
+    return null;
+  }
```
juggler434: I think it makes more sense to move the isDevisive and isHighBridging aboe this and check against them
```suggestion
  if (!forceShow && !isDevisive && !isBridging) {
    return null;
  }
```

next-client/src/components/claim/Claim.tsx
```diff
@@ -64,3 +77,49 @@ export function QuoteIcon({ num }: { num: number }) {
...
+    ? "text-green-600"
+    : isDivisive
+      ? "text-red-600"
+      : "text-gray-600";
+
+  const simpleLabel = isHighBridging
+    ? "Bridging"
+    : isDivisive
+      ? "Divisive"
+      : "Neutral";
```
juggler434: Not a fan of the nested Ternary statements. 

would rather something like 

let colorClass = neutralColor
let simpleLabel = neutral

if isHighBridging
    colorClass = greenColor
    simpleLabel = "Bridging"
if isDivisive
   colorClass = redColor
   simpleLabe = "Devisive"

Also, these colors and number don't match what you have defined in the utility file, is this intentional?

next-client/src/lib/bridging/utils.ts
```diff
@@ -0,0 +1,161 @@
+/**
+ * Utility functions for bridging-based report structuring.
+ *
+ * Provides functions to retrieve, format, and visualize bridging scores from
+ * Perspective API analysis.
+ */
+
+import type { AddOns, ClaimBridgingScore } from "tttc-common/schema";
```
juggler434: I think it would be good to set the bridging scores to constants.  They are referenced a lot.

justinstimatze:
Significant rebasing, so re-requesting review.

PR #386, opened 2025-11-21
T3C-549: Add print style sheets for export

justinstimatze:

Add print style sheet support both for printings and as an easy formatted export option.

next-client

Locally

### DIFF
--- a/next-client/public/print.css
+++ b/next-client/public/print.css
@@ -1,0 +1,731 @@
+/* Print styles for PDF export
+   This file is loaded with media="print" to avoid bundling with regular page CSS */
+
+/* ========================================
+   CSS CUSTOM PROPERTIES
+   Design tokens for print layout
+   ======================================== */
+:root {
+  /* Typography - Font sizes optimized for print readability */
+  --print-h1-size: 36pt; /* Main topic headings */
+  --print-h1-first: 40pt; /* First page report title (extra emphasis) */
+  --print-h2-size: 14pt; /* Section headings */
+  --print-h3-size: 12pt; /* Subsection headings */
+  --print-h4-size: 11pt; /* Minor headings */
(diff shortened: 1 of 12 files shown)


common/schema/index.ts
```diff
@@ -244,6 +244,463 @@ function validateNoSpeakerOverlap(data: {
...
+      inputCount++;
+      const existing = speakerMap.get(id);
+      if (existing) {
+        existing.appearsIn.add(listType);
+        existing.countByList[listType]++;
+        // Preserve the richest speaker data (prefer longer strings or those with metadata like strength scores)
+        // Examples: "1:Alice | 0.9" is richer than "1:Alice", "1:Alice Smith" is richer than "1:Alice"
+        if (
+          speaker.length > existing.fullString.length ||
+          (speaker.includes("|") && !existing.fullString.includes("|"))
```
juggler434: This could prioritize a speaker with a longer name over a speaker with additional metadata.  If we want to prioritize keeping the strength score we could do something like this:


```suggestion
if (
  (speaker.includes("|") && !existing.fullString.includes("|")) ||
  (speaker.includes("|") === existing.fullString.includes("|") && speaker.length > existing.fullString.length)
)
```

next-client/src/app/global.css
```diff
@@ -193,3 +193,704 @@
     @apply stroke-[1.5];
   }
 }
+
+/* Print styles for PDF export */
```
juggler434: This is a whole lot of extra CSS that will get bundled with every page.  Is there a way to create a separate print styles css file that can be referenced specifically when needed?

juggler434 (changes requested):
Overall looks good, some nitpicks

juggler434 (approved):
LGTM

PR #387, opened 2025-11-21
T3C-782: Fix NextJS logger dependency error

justinstimatze:

Address a logging dependency error on NextJS startup.

next-client, common

Locally


juggler434 (approved):
Thanks for fixing that

PR #417, opened 2025-12-08
T3C-729: Add monday.com CRM integration for progressive profiling

justinstimatze:

Replace old external waitlist form with sign-up flow.

- Sync new users to Monday.com at signup with upsert logic (prevents duplicates)
- Add GDPR-compliant newsletter checkbox to ProfileSetupModal (unchecked default)
- Return isNew flag from ensureUserDocument for conditional Monday sync
- Track profile completion: Company AND Use Case required for "complete" status
- Partial submissions count as skip (reprompt in 7 days, max 3 times)
- Use parameterized GraphQL queries to prevent injection attacks

next-client, express-server

Locally

### DIFF
--- a/express-server/src/routes/profile.ts
+++ b/express-server/src/routes/profile.ts
@@ -1,0 +1,174 @@
+/**
+ * User Profile Management Routes
+ *
+ * Handles progressive profiling for monday.com CRM integration.
+ * Users can update their profile information which syncs to monday.com.
+ */
+
+import { Response } from "express";
+import { RequestWithLogger } from "../types/request";
+import { z } from "zod";
+import * as firebase from "../Firebase";
+import { sendError } from "./sendError";
+import { createMondayItem } from "../services/monday";
+import { logger } from "tttc-common/logger";
+import { DecodedIdToken } from "firebase-admin/auth";
+
+const profileLogger = logger.child({ module: "profile" });
+
+/**
+ * Basic phone number validation regex
+ * Allows: digits, spaces, hyphens, parentheses, plus sign
+ * Examples: +1 (555) 123-4567, 555-123-4567, +44 20 7946 0958
+ */
+const phoneRegex = /^[+]?[\d\s\-()]{7,20}$/;
+
+/**
+ * Profile update request schema
+ * All fields are optional for progressive profiling
+ */
+const profileUpdateSchema = z.object({
+  company: z.string().trim().min(1).max(200).optional(),
+  title: z.string().trim().min(1).max(100).optional(),
+  phone: z
+    .string()
+    .trim()
+    .min(7)
+    .max(20)
+    .regex(phoneRegex, "Invalid phone number format")
+    .optional(),
+  useCase: z.string().trim().min(1).max(500).optional(),
+  newsletterOptIn: z.boolean().optional(),
+});
+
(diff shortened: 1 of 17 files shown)


express-server/src/services/monday.ts
```diff
@@ -0,0 +1,379 @@
...
+export function isMondayEnabled(): boolean {
+  return !!(process.env.MONDAY_API_TOKEN && process.env.MONDAY_BOARD_ID);
+}
+
+/**
+ * monday.com column ID mapping
+ * These IDs are specific to our monday.com board (ID: 18391158937)
+ * Update these values when moving to production board
+ */
+const COLUMN_IDS = {
```
juggler434: Should these be environment variables so we can set them differently for staging?

express-server/src/services/monday.ts
```diff
@@ -0,0 +1,379 @@
...
+  }
+  if (profile.phone) {
+    columnValues[COLUMN_IDS.phone] = profile.phone;
+  }
+  if (profile.useCase) {
+    columnValues[COLUMN_IDS.useCase] = profile.useCase;
+  }
+
+  // Newsletter opt-in checkbox
+  if (profile.newsletterOptIn !== undefined) {
```
juggler434: Is there a reason that we don't want to just store false as a default value if a user doesn't opt in?

express-server/src/routes/profile.ts
```diff
@@ -0,0 +1,149 @@
...
+import { logger } from "tttc-common/logger";
+
+const profileLogger = logger.child({ module: "profile" });
+
+/**
+ * Profile update request schema
+ * All fields are optional for progressive profiling
+ */
+const profileUpdateSchema = z.object({
+  company: z.string().min(1).max(200).optional(),
```
juggler434: for these string values a single white space will pass. 
```suggestion
  company: z.string().trim().min(1).max(200).optional(),
```

juggler434 (commented):
Some suggestions, would also be nice if there were some tests around this

next-client/src/components/navbar/components/LoginButton.tsx
```diff
@@ -26,31 +23,119 @@ import { useSearchParams } from "next/navigation";
...
+  const [showProfileSetup, setShowProfileSetup] = useState<boolean>(false);
   const searchParams = useSearchParams();
 
+  // Show profile setup modal based on skip tracking
+  // - First time: show immediately after signup
+  // - If skipped: show again after 7 days, up to 3 times
   useEffect(() => {
-    if (isWaitlisted) {
-      setShowAccessLimitedModal(true);
+    if (user && !loading) {
```
juggler434: Claude flags a possible race condition here:

```Effect runs, checks shownThisSession (null)
Effect runs again in strict mode before sessionStorage.setItem completes
Modal shown twice
Recommendation: Use a ref to track if modal has been shown:

const hasShownModal = useRef(false);

useEffect(() => {
  if (user && !loading && !hasShownModal.current) {
    const shownThisSession = sessionStorage.getItem("profileSetupShownThisSession");
    if (shownThisSession) return;

    if (shouldShowProfileSetup()) {
      hasShownModal.current = true;
      sessionStorage.setItem("profileSetupShownThisSession", "true");
      setShowProfileSetup(true);
    }
  }
}, [user, loading]);``

PR #428, opened 2025-12-10
T3C-685: Migrate Firebase Auth from popup to redirect flow

justinstimatze:

Avoid popup based Firebase Auth flow due to popup blocker interference.

next-client

Locally


justinstimatze:
Canceling this change.

PR #433, opened 2025-12-11
T3C-763 Convert Sort and Deduplicate Claims step to Pipeline worker

juggler434:
Convert the sorting/declustering step of the pipeline to the node worker
pipeline-worker
unit tests

### DIFF
--- a/pipeline-worker/src/pipeline-steps/sort-and-deduplicate/model.ts
+++ b/pipeline-worker/src/pipeline-steps/sort-and-deduplicate/model.ts
@@ -1,0 +1,180 @@
+/**
+ * Deduplication model using OpenAI API
+ */
+
+import OpenAI, { APIError, RateLimitError, BadRequestError } from "openai";
+import { Result, success, failure } from "tttc-common/functional-utils";
+import { logger } from "tttc-common/logger";
+import { initializeWeaveIfEnabled } from "../utils";
+import type {
+  Claim,
+  DeduplicationResponse,
+  LLMConfig,
+  TokenUsage,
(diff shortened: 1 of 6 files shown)


pipeline-worker/src/pipeline-steps/sort-and-deduplicate/model.ts
```diff
@@ -0,0 +1,171 @@
...
+      },
+    });
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : String(error);
+
+    // Detect specific error types for better logging and debugging
+    const isRateLimit =
+      errorMessage.toLowerCase().includes("rate limit") ||
+      errorMessage.toLowerCase().includes("429") ||
+      (error as any)?.status === 429;
```
justinstimatze: Could be an error type, also line 105, 120.

pipeline-worker/src/pipeline-steps/sort-and-deduplicate/model.ts
```diff
@@ -0,0 +1,171 @@
...
+  const content = response.output_text;
+  if (!content) {
+    dedupLogger.error({ ...context, response }, "No response from model");
+    return failure(new EmptyResponseError(llmConfig.model_name));
+  }
+
+  // Parse JSON response
+  let dedupClaims: DeduplicationResponse;
+  try {
+    const parsed = JSON.parse(content);
```
justinstimatze: Should probably be Zod validating these returns but I think that's a whole other ticket I'm about to make, just noting.

PR #457, opened 2025-12-17
T3C-765 Convert Summaries step to pipeline worker

juggler434:
Converts the Summaries step of the Pipeline.  

Currently comparing convert and sort branch until that gets merged in

Includes some other fixups

### DIFF
--- a/pipeline-worker/src/pipeline-steps/summaries/index.ts
+++ b/pipeline-worker/src/pipeline-steps/summaries/index.ts
@@ -1,0 +1,216 @@
+/**
+ * Topic Summaries Pipeline Step
+ *
+ * Given a sorted and deduplicated tree with topics, subtopics, and claims, this step:
+ * 1. Generates a summary for each topic using an LLM
+ * 2. Processes each topic individually in parallel with controlled concurrency
+ * 3. Aggregates usage and cost statistics
+ *
+ * Error handling: If any topic fails to generate a summary, the entire operation
+ * fails fast and returns the first error encountered.
+ */
+
+import OpenAI from "openai";
+import type { Logger } from "pino";
+import { failure, type Result, success } from "tttc-common/functional-utils";
+import { sanitizeForOutput } from "../sanitizer";
+import { getReportLogger, processBatchConcurrently } from "../utils";
+import { callSummaryModel } from "./model";
+import type {
+  ClusteringOptions,
+  LLMConfig,
+  SortedTree,
+  SummariesInput,
+  SummariesResult,
+  SummaryModelResult,
+  TopicSummary,
+} from "./types";
+import { ClusteringError } from "./types";
+
+/**
+ * Maximum number of topics to process concurrently to avoid rate limits
+ */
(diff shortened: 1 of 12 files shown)


pipeline-worker/src/pipeline-steps/summaries/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  SummariesResult,
+  SortedTree,
+  LLMConfig,
+  ClusteringOptions,
+  ClusteringError,
+  TopicSummary,
+  TokenUsage,
+} from "./types";
+
+const summariesLogger = logger.child({ module: "summaries" });
```
justinstimatze: Never used

pipeline-worker/src/pipeline-steps/summaries/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  );
+
+  // Check for failures
+  const firstFailure = results.find((r) => r.tag === "failure");
+  if (firstFailure && firstFailure.tag === "failure") {
+    reportLogger.error(
+      { error: firstFailure.error },
+      "At least one topic summary failed",
+    );
+    return firstFailure;
```
justinstimatze: Seems like this fails fast but the file header says that processing continues even with failures.

pipeline-worker/src/pipeline-steps/summaries/model.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  const context = {
+    topic: topicName,
+    reportId,
+  };
+
+  summaryLogger.info(context, `Generating summary for topic: ${topicName}`);
+
+  // Build prompt with tree data
+  const sanitizedPrompt = basicSanitize(userPrompt);
```
justinstimatze: Seems like you get an object back and not a string here.

pipeline-worker/src/pipeline-steps/summaries/model.ts
```diff
@@ -0,0 +1,252 @@
...
+  summary: string,
+  tree: SortedTree,
+): void {
+  const llmJudgeScorer = createLLMJudgeScorer(openaiClient);
+
+  // Run scorers on the result we already have (non-blocking)
+  // Note: The LLM judge scorer will make its own LLM call for evaluation
+  // Scores are automatically sent to Weave since scorers are wrapped with weave.op
+  Promise.all([
+    summariesJsonStructureScorer({
```
justinstimatze: Claude says:

> **Lines 52-67** (runSummaryScorers)                                                                                                                    
>   Type mismatch with scorer input types. The scorers expect Claim.duplicates to always be a defined array, but ProcessedSubtopic.claims has type Claim[] 
>   where duplicates is optional (Claim[] | undefined).                                                                                                    
>                                                                                                                                                          
>   Options:                                                                                                                                               
>   1. Update types.ts to make duplicates required in processed claims (after deduplication they should always exist, even if empty array)                 
>   2. Or cast/transform the data before passing to scorers

justinstimatze (approved):
Missed one non-blocking thing, though it works.

PR #532, opened 2025-12-28
T3C-1028: ACK Pub/Sub messages only after successful processing

justinstimatze:
## Summary

- Fix reliability issue where Pub/Sub messages were ACK'd immediately after parsing, before `processJob()` completed
- If processing failed, the message was permanently lost since Pub/Sub considered it delivered
- Now ACK only after successful processing, NACK on any failure (allowing Pub/Sub to redeliver)
- Matches the pattern already used in `pipeline-worker/src/queue/googlepubsub.ts`


juggler434:
Updated implementation to read if a message is being processed, and if so ignore it.  Saw it work as intended.

PR #538, opened 2025-12-29
T3C-966: Block oversized comments with soft warning at 5K chars

justinstimatze:
## Summary

- **Hard error at 10,000 characters**: Blocks CSV submission with actionable error showing affected IDs
- **Soft warning at 5,000 characters**: Suggests segmentation for better quality but allows proceeding
- **Defense-in-depth**: Sanitizers now reject (not truncate) oversized content if client validation is bypassed

Resolves the issue where comments exceeding 10K characters were silently truncated in the pipeline without user notification.

### DIFF
--- a/next-client/src/components/create/components/FormSections.tsx
+++ b/next-client/src/components/create/components/FormSections.tsx
@@ -242,6 +242,17 @@ export function FormDataInput({
             : `File is too large - ${formatBytes(userSizeLimit)} limit`;
         setInlineError(errorMessage);
         handleReset(inputRef);
+      } else if (result.error.tag === "Oversized Comments") {
+        // Oversized comments - show inline error with details
+        const { count, maxLength, affectedIds } = result.error;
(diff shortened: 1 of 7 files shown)


next-client/src/components/create/components/FormSections.tsx
```diff
@@ -242,6 +242,16 @@ export function FormDataInput({
             : `File is too large - ${formatBytes(userSizeLimit)} limit`;
         setInlineError(errorMessage);
         handleReset(inputRef);
+      } else if (result.error.tag === "Oversized Comments") {
+        // Oversized comments - show inline error with details
+        const { count, maxLength, affectedIds } = result.error;
+        const idsPreview = affectedIds.slice(0, 3).join(", ");
+        const moreText = affectedIds.length > 3 ? "..." : "";
```
juggler434: where is the 3 coming from?

PR #640, opened 2026-04-07
T3C-1142 Adds model select

juggler434:


express-server/src/routes/create.ts
juggler434, line 1: @claude New issue: Primitive Obsession
Check warning: ❌ New issue: Primitive Obsession
In this module, 30.6% of all function arguments are primitive types, threshold = 30.0%. The functions in this file have too many primitive types (e.g. int, double, float) in their function argument lists. Using many primitive types lead to the code smell Primitive Obsession. Avoid adding more primitive arguments.

next-client/src/components/create/CreateReport.tsx
juggler434, line 1: @claude ❌ New issue: Overall Code Complexity
Check warning: ❌ New issue: Overall Code Complexity
This module has a mean cyclomatic complexity of 4.25 across 8 functions. The mean complexity threshold is 4. This file has many conditional statements (e.g. if, for, while) across its implementation, leading to lower code health. Avoid adding more conditionals.

PR #701, opened 2026-08-06
Fix transitive dependency CVEs via pnpm.overrides

justinstimatze:
## Summary
- GitHub Dependabot alerts (150 open) are almost entirely transitive-only packages (axios, undici, protobufjs, fast-uri, node-forge, systeminformation, grpc-js, and ~40 more) that aren't declared in any `package.json`, so Renovate never opens PRs for them. Added scoped `pnpm.overrides` in root `package.json` pinning each to its patched version, scoped by major line where multiple majors are resolved in the lockfile (ajv, bn.js, brace-expansion, form-data, picomatch, grpc-js) so we don't force an incompatible major onto a dependent that pinned an older range.
- Folds in the 9 already-open, already-green Renovate PRs that were stuck because repo-level "Allow auto-merge" was off (next 16.3.0, postcss, autoprefixer, zustand, effect, fs-extra, rollup) — those PRs can be closed once this merges, no separate action needed.
- `renovate.json`: explicit `vulnerabilityAlerts.automerge` and `lockFileMaintenance` so future single-CVE and periodic transitive bumps merge themselves once green, now that auto-merge is actually enabled on the repo.

**Not touched, flagged for follow-up:**
- `elliptic` — 6.6.1 is already the latest published version; no fix exists upstream yet.
- `pm2` — direct devDependency, needs a major bump (6→7) to fix; dev-only tooling (not shipped to prod), low severity. Left for normal Renovate major-update review rather than forcing.

## Test plan
- [x] `pnpm build` — 6/6 packages build clean
- [x] `pnpm test` (with CI-matching env vars) — 6/6 packages, 1942 tests pass, 0 failures
- [x] Verified every override's resolved version in `pnpm-lock.yaml` matches its target


justinstimatze:
Re-triggering CI - PR Validation workflow never fired on the pull_request event for this branch.

